PR #58, opened 2022-02-24
Replaced usage of function<> with inplace_function<> for fixed alloca…

sherief:
…tion behavior.

This change guarantees job system lambdas won't unpredictably decide to allocate at runtime.


sherief:
CLA won't be signed. I'm glad to offer this contribution to the project under the same MIT license the original files were generously offered under, but that's as far as it'll go.

jrouwe:
I'm sorry, I cannot accept this change. I don't want to include 3rd party software in this repository (BOOST in this case), the only allowed dependency is the standard template library. Not allocating memory at run-time (i.e. during the physics step) was an important aspect during the design of the library. As such, all uses of std::function have been checked to not actually go through malloc.

Can I ask what issue you have with signing the CLA? (I think I picked the most liberal one I could find)

sherief:
No worries wrt closing the PR. I'd never touch Boost with a ten-foot-pole though, this file is from C++'s WG21, it just uses the Boost license: https://github.com/WG21-SG14/SG14/blob/master/SG14/inplace_function.h

I'm curious to know how you verified std::function<> won't allocate - on all libc++ implementations and MSVC STL? It feels like an implementation detail that's very easy to accidentally violate and suddenly you have a frame time spike. Again, merely curious, I totally understand the rationale for refusing this PR.

jrouwe:
I looked up the low level memalloc function for the platform and put a breakpoint there, then I ran the simulation and verified that it didn't hit it (actually I think there's 1 allocation left in PhysicsSystem::Update during the broad phase rebuild that I couldn't easily get rid of). I agree that it's not ideal.

PR #59, opened 2022-02-24
Fix compile error when lzcnt isn't enabled

mrezai:
It seems _BitScanForward and _BitScanReverse are available only in Visual Studio.


jrouwe:
Thank you for your contribution!

PR #61, opened 2022-02-25
Adds support for MacOS with UnitTests passing

fhoenig:
Makes UnitTests work for me on MacOS. I did not touch the build system and used a local premake file instead.
The usual #ifdef salad could be cleaned up a bit but it seems to work.


jrouwe:
Thank you for your contribution! I hope my changes to make it merge in and do the ifdefs in a slightly different way didn't break it.

PR #65, opened 2022-02-26
Created BroadPhaseLayerInterface so that we don't need an array mapping ObjectLayer -> BroadPhaseLayer anymore

jrouwe:


jankrassnigg (approved):
Looks great!

PR #82, opened 2022-03-08
Add SetSensor

Slartibarty:
Allows bodies to become sensors after creation, eg. bodies that switch between fluid state.

After inspecting the code this seems to be legal, and works for what we want. I'm not sure if I'm missing something important, so it'd be nice to have some verification that this doesn't adversely affect anything.


jrouwe:
The SetSensor(false) branch is missing a ~ (you're now cleaning all the other flags). Also, I'll need to check if this works correctly with the contact cache.

Slartibarty:
Alright, amended the commit. I double checked against other methods in the file that manipulate the bitflags to keep them all similar.

jrouwe:
I've looked at the code and there is a (subtle) problem that can be introduced: In BodyManager::CreateBody we assert that the body is static if it is a sensor. Through this accessor you will now be able to bypass that check, creating a dynamic sensor, which is not something that is supported (and wouldn't make sense either as it would fall through everything). Is this a problem for you?

Joshua-Ashton:
Yeah, we do not need dynamic sensors. We just need to transition already existing static bodies (in this case, world geometry) to sensors and back on-demand.

Perhaps we can add an assert to SetSensor that the object is static to solve that edge case.

jrouwe:
Yes, I'll make those changes (want to clean up creation at the same time).

PR #83, opened 2022-03-08
Add GetCenterOfMassTransform to BodyInterface

Joshua-Ashton:
This is useful for avoiding having to do the the rotation * pShape->GetCenterOfMass() on the other side, which is just inverting work GetWorldTransform is doing as it calls GetPosition and doesn't use mPosition directly.


jrouwe:
Thank you again!

PR #86, opened 2022-03-11
Fix VehicleCollisionTesters colliding with sensor bodies

Joshua-Ashton:
Fixes vehicles being able to drive on water, or collide with triggers etc.


jrouwe:
Thank you! The change is correct, but I want to point out that you probably want to catch this at a higher level too. What we do is to set the collision layers up in such a way that the vehicle raycast layer doesn't collide with any water / triggers. Catching it at this point means that you will do more work than strictly needed.

Joshua-Ashton:
Makes sense

PR #89, opened 2022-03-12
Add the ability to directly specify local reference frames for SixDOF constraints

Joshua-Ashton:
Avoids having to do a bunch of useless math if you are already working in local space.

Our existing code gives us the information for ragdoll constraints already in local space, so it seems fairly annoying and entirely redundant to extract all of the information like COM from the body, work out all the axes, for Jolt to just... go back to what we had in the first place.


jrouwe:
Hello, you're the second one to suggest this so I guess there's demand for specifying constraints in local space. I think we can solve it though without adding a completely disjoint set of variables. I've created #90 to show how I think it should be done (I implemented a number of constraint types, need to do the rest still).

If you have your local constraint matrices lc1 and lc2 then you would do:

`SixDOFConstraintSettings settings;
settings.mSpace = EConstraintSpace::LocalToBody;
settings.mPosition1 = lc1.GetTranslation();
settings.mAxisX1 = lc1.GetAxisX();
settings.mAxisY1 = lc1.GetAxisY();
settings.mPosition2 = lc2.GetTranslation();
settings.mAxisX2 = lc2.GetAxisX();
settings.mAxisY2 = lc2.GetAxisY();`

Let me know if this works for you.

jrouwe:
Changes have been committed. Closing this pull request.

PR #93, opened 2022-03-14
Add SSE2 codepath

jfpoole:
Add an SSE2 codepath to Jolt by providing a C++ implementation for all
functions that use SSE4.1 instructions.  This enables Jolt to run on
all x64 processors.

The SSE2 codepath is ~10% slower than the SSE4.2 codepath on average in
the PerformanceTest tests.


Jolt/Math/UVec4.inl
```diff
@@ -105,23 +105,31 @@ UVec4 UVec4::sGatherInt4(const uint32 *inBase, UVec4Arg inOffsets)
...
 {
-#if defined(JPH_USE_SSE)
+#if defined(JPH_USE_SSE4_1)
 	return _mm_min_epu32(inV1.mValue, inV2.mValue);
 #elif defined(JPH_USE_NEON)
 	return vminq_u32(inV1.mValue, inV2.mValue);
 #else
-	#error Unsupported CPU architecture
+	UVec4 result;
+	for (int i = 0; i < 4; i++) {
```
jrouwe, line 114: Jolt uses either:

    for (...)
    {
         multiple;
         statements;
    }

or

    for (....)
      single_statement;

Could you please remove the brackets? (here and also for the other functions you changed)

Jolt/Math/UVec4.inl
```diff
@@ -105,23 +105,31 @@ UVec4 UVec4::sGatherInt4(const uint32 *inBase, UVec4Arg inOffsets)
...
+		result.mU32[i] = std::min(inV1.mU32[i], inV2.mU32[i]);
```
jrouwe, line 115: Inside the namespace JPH we're already `using namespace std`, can you remove `std::` (2x in this file)

Jolt/Math/Vec3.inl
```diff
@@ -527,67 +533,87 @@ Vec3 Vec3::Cross(Vec3Arg inV2) const
...
 	mul = vsetq_lane_f32(0, mul, 3);
     float32x2_t sum = vdup_n_f32(vaddvq_f32(mul));
     return vget_lane_f32(vsqrt_f32(sum), 0);
 #else
-	#error Unsupported CPU architecture
+	float len_sq = 0.0f;
+	for (int i = 0; i < 3; i++) {
+		len_sq += mF32[i] * mF32[i];
+	}
+	return sqrtf(len_sq);
```
jrouwe, line 616: Jolt uses the C++ sqrt function instead of the C sqrtf function. Can you change that? (happens multiple times)

jrouwe:
Hello, thank you for your contribution! I added a few minor remarks (I can fix them as well if you want). Could you please sign the Contributor License Agreement (button at the top of the file)?

jfpoole:
My pleasure!  I believe I've addressed all of your comments in the latest commit.  Could you let me know if you have any additional comments?  As for the CLA, I'm just waiting to hear back from our lawyer to make sure he doesn't have any concerns.

jfpoole:
I've signed the CLA.  Let me know if there's anything else you need for this pull request.

Also, is there a style guide for the project?  I'd like to make sure any future PRs conform to the style guide.

jrouwe:
Thanks! I've merged the changes. Unfortunately I wasn't able to edit the pull request directly (you must have disabled this, so far I've always been able to do so) so I merged it as is and then made some changes on top.

Unfortunately there's no style guide for the project yet. I tried to make a clang format file for it, but it completely destroyed the indentation of most of the files. I'll write something down in the future.

jankrassnigg:
Honestly, the indentation is in my opinion completely bonkers 🤣 sorry, but I find it really not helpful as it is and in all text editors that I use it looks entirely messed up. I guess you use tabs and assume a certain tab width. Anyway, I and all my colleagues have let go of such handcrafted code layout. Use clang format,  choose one of the common styles, tweak a few details to your liking and then just always accept what it gives you. After a few weeks you'll notice that you have much more piece of mind and just don't waste any thought on this anymore.

We even set its up a github action to automatically apply code formating on PRs.

Wunkolo:
+1 for clang-format!
Would make it much easier to automate a consistent style and would remove a lot of the code-style fixup overhead from contributions.

jfpoole:
It took us a while to get our .clang-format file just right for our internal project.  @jrouwe, if you're interested, I'd be happy to try and write one that works well with this project.  I'd just need to know which files are exemplars for your preferred code style.

jrouwe:
I hope that all files are examplar :) You could e.g. pick Body.cpp/h/inl. The main problem I had was that clang format doesn't like the column approach that I'm using in header files where all member functions return values are in one column, the function names start in the next and then the inline function implementations are in a 3rd column. The variables align in the same way with 1st column being type, 2nd name and 3rd the comment. The width of the columns varies from file to file depending on the contents of the columns (although sometimes, for very long type definitions I allow them to break the rule).

PR #99, opened 2022-03-19
Allow disabling the Windows only targets.

All8Up:
This is a trivial change.  In order to build the absolute minimal needed, this allows removal of the Windows only targets.


jrouwe:
Thank you for your suggestion! I've split up the TARGET_WINDOWS_ONLY into a TARGET_SAMPLES and TARGET_VIEWER to have a bit more fine grained control.

PR #102, opened 2022-03-23
CMake FetchContent Support

tksuoran:
This adds a minimal CMake `CMakeLists.txt` so that Jolt Physics can be used nicely with CMake FetchContent.

Here is an example how one would use CMake FetchContent:

```
include(FetchContent)
FetchContent_Declare(
    JoltPhysics
    GIT_REPOSITORY  https://github.com/jrouwe/JoltPhysics.git
    GIT_TAG         master
    GIT_SHALLOW     TRUE
    GIT_PROGRESS    TRUE
)
FetchContent_MakeAvailable(JoltPhysics)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,5 @@
+cmake_minimum_required(VERSION 3.0.0)
+project(Jolt)
+set(PHYSICS_REPO_ROOT ${PROJECT_SOURCE_DIR})
+set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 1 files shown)


tksuoran:
Ok I was a bit silly, there actually is CMakeLists.txt in the Build directory. It is possible to use this with FetchContent, but it requires a bit more extra to handle the location not being where CMake FetchContent looks by default (in the root directory). Feel free to close.

tksuoran:
Closing as this is not needed.

PR #103, opened 2022-03-24
Fix JPH_ASSERT() macro

tksuoran:
JPH_ASSERT() macro uses identifiers from JPH namespace, and would only work with sufficient using statements without this fix.


tksuoran:
I run into this because I had copy-pasted sample code which used `JPH_ASSERT()`. I can easily replace this with my own assert.

tksuoran:
This PR is not needed. Also, I pushed unrelated work to my dev branch.

PR #109, opened 2022-03-29
Add a helper to manually invalidate contact cache

Joshua-Ashton:
In our game, sometimes an object can be forced up against a wall by a force and then a hole is cut into the wall it is on, and it doesn't update as it still thinks its it's making contact, even when forcing an Activate, etc.

We know this ahead of time so we can manually invalidate the contact cache for the object that was touching.


Joshua-Ashton:
I should add more context: the object's return value of `ShouldCollide` from the collision solver changes, as the original body (the world) will return false with it, and the new wall with the whole cut, will return true for it now.

jrouwe:
Thanks!

PR #114, opened 2022-04-02
Add Set/GetMaxFriction to SixDOFConstraint

Joshua-Ashton:
Useful if you want to change this dynamically, or are retrofitting an API where this is specified after constraint creation.


jrouwe:
There were two flags that were not properly updated. So if you created the constraint with no friction and then enabled friction later it there wouldn't be any friction. (There was a bug before that if you created the constraint with friction but never activated any motor that the friction wouldn't be active). I've fixed both.

Joshua-Ashton:
Thanks!

PR #122, opened 2022-04-08
Created a virtual character class (implemented using collision detection functions only for more fine grained character control)

jrouwe:


jankrassnigg:
Haven't tried it yet, but the interface looks very promising!

PR #156, opened 2022-05-11
Add TrackContactsForThisBodyPair to ValidateResult

Joshua-Ashton:
Adds a new validate result, in which the object's collision can be tracked for contact, without actually performing any constraints, etc on the object.

This essentially makes it work like a sensor, following the validation step, which can be useful if the body only wants to act as a sensor to some objects and not others.

It's also useful if your game/engine expects to get Start/EndTouch callbacks even when collision is disabled from an object from testing if two narrow-phase bodies should collide.

### DIFF
--- a/HelloWorld/HelloWorld.cpp
+++ b/HelloWorld/HelloWorld.cpp
@@ -163,17 +163,17 @@ public:
 		return ValidateResult::AcceptAllContactsForThisBodyPair;
 	}
 
-	virtual void			OnContactAdded(const Body &inBody1, const Body &inBody2, const ContactManifold &inManifold, ContactSettings &ioSettings) override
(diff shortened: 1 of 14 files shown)


jrouwe:
Hello, I think this is a great idea! I think it can be done in a less intrusive way though by moving it to the Add/Persist contact callbacks instead of using the Validate callback.

I've typed in what I think the proper solution would be (this also fixes a bug where persisted contacts with sensors would make the bodies end up in the same island):

https://github.com/jrouwe/JoltPhysics/tree/feature/on_demand_sensor

See commit:

https://github.com/jrouwe/JoltPhysics/commit/e9c2be39f89b03704737b2271d1cf00a01dea2bc

What do you think?

Joshua-Ashton:
Sounds great and works for us :)

jrouwe:
Merged #158 into master. Thanks for the idea! (I credited you in the change)

PR #157, opened 2022-05-11
Fix error in doctest.h under MSVC 2022

Joshua-Ashton:
Warns about not having additional inner braces for unions, however, it is easier to just {} and avoid it.


jrouwe:
Hello,

It compiles fine for me under VS2022, but the code is equivalent, so I'll merge the CL.

PR #160, opened 2022-05-13
Implemented a function to do stair walking for the virtual character

jrouwe:


Jolt/Physics/Character/CharacterVirtual.h
```diff
@@ -123,6 +123,23 @@ class CharacterVirtual : public CharacterBase
 	/// @param inAllocator An allocator for temporary allocations. All memory will be freed by the time this function returns.
 	void								Update(float inDeltaTime, Vec3Arg inGravity, const BroadPhaseLayerFilter &inBroadPhaseLayerFilter, const ObjectLayerFilter &inObjectLayerFilter, const BodyFilter &inBodyFilter, TempAllocator &inAllocator);
 
+	/// This function will return true if the character has moved into a slope that it too steep (e.g. a vertical wall).
```
jankrassnigg, line 126: minor typo here

jankrassnigg: "that IS too steep"

jrouwe: Thanks, you're quick!

jankrassnigg:
One potential addition would be to also have a "step down" vector. What I currently do is not only step OVER obstacles and stair steps, but also if the ground is within a distance of ~20cm step the character DOWN to keep touching it. This is very useful when you walk down stairs or a slope, to keep the character touching the ground instead of having it "fall" down, which often leaves the character for a few frames in a state of "not touching the ground" which is quite annoying.

I think the same logic could be handled in here, since after casting up and forwards, you just need to cast down the step-up amount plus an additional step-down amount.

jankrassnigg:
Yes, that's what I meant. It's also more explicit to users, so if they don't want step-down, it's clear to just set that to zero.

jankrassnigg:
Oh I just remembered a caveat. You only want to step down if you hit ground within that step down distance. If you hit nothing, you want to stay at the original height. Otherwise if you step over a ledge, you immediately step down in that frame, before it is known that you are now falling.
I hope that isn't too specific a feature request? Otherwise I'll deal with it myself.

jrouwe:
Currently, if there's no floor after the step up then the walk stairs is cancelled completely (normally you should hit the top of the step after your step up).

PR #161, opened 2022-05-13
Collection of misc. fixes from porting to a new platform

Joshua-Ashton:


jrouwe:
It looks like there is quite a bit of overlap with your other pull request (#162). Is this a different platform?

Joshua-Ashton:
Sorry to be unclear, this is the same one, just parts that I thought were ready for review

Jolt/Core/LockFreeHashMap.inl
```diff
@@ -114,15 +114,16 @@ void LockFreeHashMap<Key, Value>::Init(uint32 inMaxBuckets)
 	mNumBuckets = inMaxBuckets;
 	mMaxBuckets = inMaxBuckets;
 
-	mBuckets = new atomic<uint32> [inMaxBuckets];
+	constexpr size_t bucket_alignment = max<size_t>(alignof(atomic<uint32>), 16u);
```
jrouwe, line 117: I assume this is needed because otherwise Clear() complains that mBuckets is 4 byte instead of 16 byte aligned? 

Note that the Clear() function already statically asserts that sizeof atomic<uint32> == 4, so I don't think we need the bucket_alignment calculation.

AlignedAlloc(AlignUp(...)) should be replaced with AlignedAlloc(...) because AlignUp is done internally already, see Memory.cpp. If _aligned_malloc on 32 bit doesn't do that then we need to add it there or else we'll have issues in multiple places.

And can you change atomic<uint32>* -> atomic<uint32> * (add space before *)?

So all in all I think this line should be:

```mBuckets = reinterpret_cast<atomic<uint32> *>(AlignedAlloc(inMaxBuckets * sizeof(atomic<uint32>), 16);```

jrouwe:
I added 1 remark. If you change that I'll merge this PR.

Jolt/Core/LockFreeHashMap.inl
Joshua-Ashton: Done.

PR #162, opened 2022-05-13
WIP: Support for 32-bit platforms

Joshua-Ashton:
I've been writing an implementation of the Source Engine's physics library (VPhysics which uses IVP/Havok) used in games such as Half-Life 2, Portal 2, L4D2, Garry's Mod, etc with Jolt.

There are some engine branches that are 64-bit and 32-bit, and I would like to provide support for both.
I also have a few friends who are integrating Jolt into older games (such as Quake 2) or mods (for games such as Half-Life 1), most of which are 32-bit, to learn more about it.

I completely understand if you don't want to deal with having this upstream or at least being officially supported.

Leaving this as a draft as it is not complete yet, but I figured I'd start the conversation now.
As of me making this, it can at least run the samples + game physics (albeit with some complaints in Debug).

### DIFF
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -42,9 +42,13 @@
 #endif
 
 // Detect CPU architecture
-#if defined(__x86_64__) || defined(_M_X64)
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
 	// X86 CPU architecture
-	#define JPH_CPU_X64
+	#if defined(__x86_64__) || defined(_M_X64)
+		#define JPH_CPU_X64
+	#else
+		#define JPH_CPU_X86
+	#endif
 	#define JPH_USE_SSE
 
 	// Detect enabled instruction sets
@@ -158,7 +162,8 @@
 	JPH_MSVC_SUPPRESS_WARNING(5045) /* Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified */ \
 	JPH_MSVC_SUPPRESS_WARNING(4583) /* 'X': destructor is not implicitly called */				\
 	JPH_MSVC_SUPPRESS_WARNING(4582) /* 'X': constructor is not implicitly called */				\
-	JPH_MSVC_SUPPRESS_WARNING(5219) /* implicit conversion from 'X' to 'Y', possible loss of data  */
+	JPH_MSVC_SUPPRESS_WARNING(5219) /* implicit conversion from 'X' to 'Y', possible loss of data  */ \
+	JPH_MSVC_SUPPRESS_WARNING(4826) /* Conversion from 'X *' to 'JPH::uint64' is sign-extended. This may cause unexpected runtime behavior. (32-bit) */
 
 // OS-specific includes
 #if defined(JPH_PLATFORM_WINDOWS)
@@ -174,7 +179,7 @@
 	#include <limits.h>
 	#include <string.h>
 
-	#if defined(JPH_CPU_X64)
+	#if defined(JPH_CPU_X86) || defined(JPH_CPU_X64)
 		#define JPH_BREAKPOINT		__asm volatile ("int $0x3")
 	#elif defined(JPH_CPU_ARM64)
 		#define JPH_BREAKPOINT		__builtin_trap()
@@ -240,7 +245,7 @@ static_assert(sizeof(uint8) == 1, "Invalid size of uint8");
 static_assert(sizeof(uint16) == 2, "Invalid size of uint16");
 static_assert(sizeof(uint32) == 4, "Invalid size of uint32");
 static_assert(sizeof(uint64) == 8, "Invalid size of uint64");
-static_assert(sizeof(void *) == 8, "Invalid size of pointer");
+static_assert(sizeof(void *) == 8 || sizeof(void *) == 4, "Invalid size of pointer" );
 
 // Define inline macro
 #if defined(JPH_COMPILER_CLANG) || defined(JPH_COMPILER_GCC)
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -44,10 +44,11 @@
 // Detect CPU architecture
 #if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
 	// X86 CPU architecture
+	#define JPH_CPU_X86
 	#if defined(__x86_64__) || defined(_M_X64)
-		#define JPH_CPU_X64
+		#define JPH_CPU_ADDRESS_BITS 64
 	#else
-		#define JPH_CPU_X86
+		#define JPH_CPU_ADDRESS_BITS 32
 	#endif
 	#define JPH_USE_SSE
 
@@ -88,6 +89,7 @@
 	// ARM64 CPU architecture
 	#define JPH_CPU_ARM64
 	#define JPH_USE_NEON
+	#define JPH_CPU_ADDRESS_BITS 64
 #else
 	#error Unsupported CPU architecture
 #endif
@@ -179,7 +181,7 @@
 	#include <limits.h>
 	#include <string.h>
 
-	#if defined(JPH_CPU_X86) || defined(JPH_CPU_X64)
+	#if defined(JPH_CPU_X86)
 		#define JPH_BREAKPOINT		__asm volatile ("int $0x3")
 	#elif defined(JPH_CPU_ARM64)
 		#define JPH_BREAKPOINT		__builtin_trap()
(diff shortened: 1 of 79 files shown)


jrouwe:
Actually it looks a lot less bad that I thought it would. There are some little things that I would like to change (mostly code style related), but if you give me edit access (if you haven't already) I can fix that. Also I can at least add e.g. the 32 bit Debug version to the builds so that it will stay working. Let me know when you're done with it.

Joshua-Ashton:
Will let you know soon, once I fix all of the issues in Debug

Jolt/Core/Core.h
```diff
@@ -42,9 +42,13 @@
 #endif
 
 // Detect CPU architecture
-#if defined(__x86_64__) || defined(_M_X64)
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
 	// X86 CPU architecture
-	#define JPH_CPU_X64
```
jrouwe, line 47: I think we should probably just rename JPH_CPU_X64 to JPH_CPU_X86 and have a separate define for JPH_CPU_ADDRESS_BITS (= 32 or 64, not entirely happy with the name but I can't come up with a better one)

Jolt/Core/Core.h
```diff
@@ -240,7 +245,7 @@ static_assert(sizeof(uint8) == 1, "Invalid size of uint8");
 static_assert(sizeof(uint16) == 2, "Invalid size of uint16");
 static_assert(sizeof(uint32) == 4, "Invalid size of uint32");
 static_assert(sizeof(uint64) == 8, "Invalid size of uint64");
-static_assert(sizeof(void *) == 8, "Invalid size of pointer");
+static_assert(sizeof(void *) == 8 || sizeof(void *) == 4, "Invalid size of pointer" );
```
jrouwe, line 248: This can then be #if JPH_CPU_ADDRESS_BITS == 32 check that pointer size is 4, else check that it is 8

Jolt/Physics/Body/Body.h
```diff
@@ -306,7 +306,11 @@ class Body : public NonCopyable
 	EMotionType				mMotionType;													///< Type of motion (static, dynamic or kinematic)
 	atomic<uint8>			mFlags = 0;														///< See EFlags for possible flags
 	
-	// 121 bytes up to here
+	// 121 bytes up to here (64-bit)
+#if UINTPTR_MAX == UINT32_MAX
```
jrouwe, line 310: this can use JPH_CPU_ADDRESS_BITS too

Jolt/Physics/Body/Body.h
```diff
@@ -306,7 +306,11 @@ class Body : public NonCopyable
...
+	// 32-bit padding
+	char padding[19];
```
jrouwe, line 312: Should be mPadding and properly tabbed.

Jolt/Physics/Collision/Shape/CompoundShape.h
```diff
@@ -226,7 +226,7 @@ class CompoundShape : public Shape
 		// 3 padding bytes left
 	};
 
-	static_assert(sizeof(SubShape) == 40, "Compiler added unexpected padding");
+	static_assert(sizeof(SubShape) == (sizeof(void*) == 8 ? 40 : 36), "Compiler added unexpected padding");
```
jrouwe, line 229: Can use JPH_CPU_ADDRESS_BITS too

jrouwe:
I tried running this myself (and in the process made a batch file to produce a solution which I just submitted). I see there's still lots of calling convention issues left.

Joshua-Ashton:
> I see there's still lots of calling convention issues left.

Yeah, am aware of those. Just made the PR now to start the conversation.

Jolt/Core/Core.h
Joshua-Ashton: What are your thoughts on only having JPH_CPU_X86 and using `sizeof(void*) == X` + ternaries  given it's consteval?

I myself am not a big macro fan, but if you'd prefer that we can do that.

jrouwe: Since all platform specifics are already done using defines, I'd keep it a define for consistency.

jrouwe:
The calling convention issues can be fixed as in this change:

https://github.com/jrouwe/JoltPhysics/commit/f3535cbe457528e9efa027d906dd0739de278e0b

I created a branch here:

https://github.com/jrouwe/JoltPhysics/tree/experiment/calling_convention_fix

Feel free to merge it into your own branch (I didn't want to interfere with anything you're doing).

The only thing that is still crashing is the Virtual Character test, probably I am using a Vec3 as argument instead of a Vec3Arg somewhere, I'll take a quick look.

This is not the most efficient fix as any vector will get passed by reference (except when stuff is inlined - which is the case for most performance critical stuff). An alternative would be to force the __vectorcall calling convention for all functions, I think there's a compiler flag for that but I haven't looked into it yet.

Joshua-Ashton:
Thanks, will take a look. I have some ideas that shouldn't need to do this, will try when at PC :)

jrouwe:
CharacterVirtualTest is also fixed. I really don't understand why MS doesn't just by default allocate objects on 16 byte boundaries in Win32, see:

https://github.com/jrouwe/JoltPhysics/commit/4d01406d041037c1d94c39e257c67cea5023de8f

Joshua-Ashton:
`This is not the most efficient fix as any vector will get passed by reference (except when stuff is inlined - which is the case for most performance critical stuff)`

This is happening anyway on your 64-bit builds right now despite passing the Vec3 type by value anyway due to the default x64 calling convention in MSVC anyway though afaiu?

https://godbolt.org/z/Y7TsGshca

Joshua-Ashton:
See here that in win64, passing by value and ref are the same: https://godbolt.org/z/GjcGcW3bj

jrouwe:
It's been years since I have looked at this stuff. Back then I had __vectorcall all over the place and I really disliked it. At some point I concluded that it was not needed anymore because the compiler would pass things in registers anyway (I think I was using /Gv at that time to force __vectorcall everywhere, but I don't remember when/why I removed this). Also this:

https://docs.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-170#register-volatility-and-preservation

clearly explains that unless you use __vectorcall, things will be passed through the stack instead of via a register.

Note that passing const Vec3 & or Vec3 is not the same as you can see here:

https://godbolt.org/z/MWd1v7zPM

In the first case it will pass the address to the original vector (which may live in a different cache page) and in the second case it will copy it to the local stack first.

In any case, I'll put it on my todo list to look into this again. For now, I think on x86 we should just use const X & for everything to avoid the issues (please leave it to be passed by value on other platforms).

Joshua-Ashton:
> clearly explains that unless you use __vectorcall, things will be passed through the stack instead of via a register.

Yes, the default on Windows sucks even on 64-bit as usual :b

On Linux, (w/ sysv amd64 abi) it just does what you want and expect which is passing it by xmm0, etc.
https://godbolt.org/z/sE5KnP481

> In the first case it will pass the address to the original vector (which may live in a different cache page) and in the second case it will copy it to the local stack first.

Would that really provide any cache benefit given you are just copying the same value you would pass by real ref otherwise to the stack just before anyway?

> For now, I think on x86 we should just use const X & for everything to avoid the issues (please leave it to be passed by value on other platforms).

Can we really do that though?
What about aliasing problems like the following?:

https://godbolt.org/z/heaWrf4ea

Is the entire codebase clear of cases like that?

jrouwe:
> Would that really provide any cache benefit given you are just copying the same value you would pass by real ref otherwise to the stack just before anyway?

Depends. Not much probably. But I'd like to look at reintroducing __vectorcall again to make sure we don't copy via stack.

> Ie. does it ever return a reference to a vec3 or other type to the application or use that internally somewhere?

It is possible of course, but all math functions return by value rather than a reference. I don't expect any aliasing issues.

jrouwe:
I managed to fix the calling convention issues by explicitly adding an alignas(16) to the Vec3/Quat etc. classes. I think it is a compiler bug and has nothing to do with __vectorcall. The caller was passing the address of a Quat and the function then tried to read it as a Quat by value (both on the stack).

jrouwe:
I merged the change, but there is 1 issue left: In Release/Distribution mode there is 1 unit test (TestQuatConvertMatrix) that fails. I think this is a compiler issue too as the code produces a matrix with a non-zero element for the last element of the first column of the matrix. It's really hard to debug as when you start modifying the code / delete other unit tests, the failure case goes away.

Joshua-Ashton:
Thanks for taking a look and finishing it off!

jrouwe:
No problem! Send me a video / link when those games are working. I think it's a cool project!

Admer456:
I'd like to express my sincere gratitude for you both, as I've been meaning to try out JoltPhysics in the context of a Half-Life 1 mod. I ran into a roadblock when I discovered it was 64-bit only, then a friend of mine pointed me to this PR hours after it was made. [Here](https://twitter.com/theAdmer456/status/1525555667939078145)'s the progress so far, and I gotta say I'm super happy with JoltPhysics.

PR #198, opened 2022-07-04
Adds optional body filter to DrawBodies()

DocAce:


Jolt/Physics/Body/BodyManager.cpp
```diff
@@ -558,14 +559,14 @@ bool BodyManager::RestoreState(StateRecorder &inStream)
 }
 
 #ifdef JPH_DEBUG_RENDERER
-void BodyManager::Draw(const DrawSettings &inDrawSettings, const PhysicsSettings &inPhysicsSettings, DebugRenderer *inRenderer)
+void BodyManager::Draw(const DrawSettings &inDrawSettings, const PhysicsSettings &inPhysicsSettings, DebugRenderer *inRenderer, const BodyDrawFilter* inBodyFilter)
```
jrouwe, line 562: Can you please write this as const BodyDrawFilter *inBodyFilter (shift space), I know it's weird, but it's unfortunately the style I adopted 10+ yrs ago.

Jolt/Physics/PhysicsSystem.h
```diff
@@ -114,7 +114,7 @@ class PhysicsSystem : public NonCopyable
 	static bool					sDrawMotionQualityLinearCast;								///< Draw debug info for objects that perform continuous collision detection through the linear cast motion quality
 
 	/// Draw the state of the bodies (debugging purposes)
-	void						DrawBodies(const BodyManager::DrawSettings &inSettings, DebugRenderer *inRenderer) { mBodyManager.Draw(inSettings, mPhysicsSettings, inRenderer); }
+	void						DrawBodies(const BodyManager::DrawSettings &inSettings, DebugRenderer *inRenderer, const BodyDrawFilter *inBodyFilter = nullptr) { mBodyManager.Draw(inSettings, mPhysicsSettings, inRenderer); }
```
jrouwe, line 117: You forgot to pass inBodyFilter to the body manager.

DocAce: Oops, how embarassing! Because of the way our project is set up I can't commit my changes from the code I tested this in so I had to manually re-apply them to my Github checkout, and I only tested the case of not passing in a filter there... I have to test more thoroughly next time.

jrouwe:
No, the reason it's failing is that github actions run under the account of who is pushing code changes. Sonar requires an access token that is in my secrets list and not in yours so it fails. This is for security reasons so that you can't check in an echo $SECRET_KEY in a build script to obtain the token. I haven't found a way to fix Sonar yet.

PR #207, opened 2022-07-23
Add MinGW cross compile

mrezai:
This setup can be used to run samples on Linux using wine and vkd3d.
Tested on Ubuntu 22.04 with wine-7.13 (Staging).

### DIFF
--- a/Build/README.md
+++ b/Build/README.md
@@ -85,7 +85,7 @@ To implement your custom memory allocator override Allocate, Free, AlignedAlloca
 
 - Follow download instructions for MSYS2 (https://www.msys2.org/)
 - From the MSYS2 MSYS app run: pacman -S --needed mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake
-- From the MSYS2 MINGW x64 app, in the Build folder run: ./cmake_mingw.sh
+- From the MSYS2 MINGW x64 app, in the Build folder run: ./cmake_windows_mingw.sh
 - Run: cmake --build MinGW_Debug
 - Run: MinGW_Debug/UnitTests.exe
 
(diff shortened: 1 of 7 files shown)


jrouwe:
B.t.w. I tested this on Ubuntu 22.04 running on Virtual Box on Windows 11. I think the Vulkan drivers were properly installed (vulkaninfo showed an NVIDIA driver among some other drivers) but as soon as I start Samples.exe it errors out with a message that it can't compile the shaders. Any tips?

jrouwe:
Thanks for the change!

mrezai:
What is the version of wine?
test it like this to see if it make any difference:
`gamemoderun /usr/bin/wine Samples.exe`

jrouwe:
Nope, same error "Failed to compile vertex shader". My wine version is 6.0.3.
Anyway I think the issue is in the Vulkan driver. It looks like it is using the 'lavapipe' implementation. I'll play around with this some more.

mrezai:
it seems Wine 7.4 is the first version that bundled vkd3d(I tested with Wine 7.13 Staging):
https://www.phoronix.com/news/Wine-7.4-Released

Maybe another option is installing Lutris to install and configure different version of Wine:
https://lutris.net/downloads
```
sudo add-apt-repository ppa:lutris-team/lutris
sudo apt update
sudo apt install lutris
```
Add Samples.exe as a game and run it with one of installed Wine.

jrouwe:
Thanks! I installed the latest develop version and now it's much clearer in its error message. It's definitively not finding the GPU and I think Virtual Box is the issue (the internet says that it doesn't allow GPU passthrough), so I'm guessing I have to either setup Linux for dual boot or try HyperV. That's for another day...

PR #219, opened 2022-08-29
Suppress `-Wunqualified-std-cast-call` for clang 16 or later

JJL772:
`-Wunqualified-std-cast-call` warns on missing qualifiers to `std::` functions, ie `std::` missing from `std::move`, even if the compiler can resolve `move(x)` properly. 
This warning flag was added in clang 16.

There are quite a few instances of unqualified `move` calls in Jolt and frankly I have no idea why clang even added this warning, so I just decided to ignore it rather than "fixing" them.


jrouwe:
Thank you!

PR #240, opened 2022-09-27
Bugfix: Kinematic sensor vs kinematic object would not register contact depending on order of activation

jrouwe:


jankrassnigg:
:-)

PR #246, opened 2022-10-08
Added CMake option for which MSVC runtime library to use

mihe:
I ran into issues with mismatching MSVC runtime libraries (error LNK2038) while trying to link against Jolt and was hoping you'd be open to adding a new CMake option to toggle between static and DLL runtime, similar to [what GLFW does](https://github.com/glfw/glfw/blob/dd8a678a66f1967372e5a5e3deac41ebf65ee127/CMakeLists.txt#L46-L47).

I moved the existing UWP check into the option itself using `cmake_dependent_option`, which means that the option won't be available at all if you're targeting UWP or if you're compiling with something other than MSVC.


jrouwe:
Thank you!

PR #253, opened 2022-10-09
Expose PointConstraint points

tksuoran:
This is just a proof of concept / reference for something that I needed but could not figure out how to do.

As is, does not properly handle different constraint space settings.


jrouwe:
I've updated the functions to include the constraint space parameter.

PR #260, opened 2022-10-10
Fix unqualifed usage of std::move and std::forward

tksuoran:
At least Clang 15.0.2 complains about unqualified use of 'std::move' and 'std::forward'.


jrouwe:
See comment in #261, this PR should no longer be needed.

Issue #261, opened 2022-10-10
Compilation errors with Clang 15.0.2


tksuoran:
Trying to compile Jolt Physics with Clang 15.0.2 gives plenty of compilation errors.
```
$ clang-15 --version
Ubuntu clang version 15.0.2-++20221003113142+5f620d003fda-1~exp1~20221003113228.69
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
Fix proposal in #260 
```
[main] Building folder: JoltPhysics 
[build] Starting build
[proc] Executing command: /usr/bin/cmake --build /home/tksuoran/git/tksuoran/JoltPhysics/build --config Debug --target all --
[build] [4/179   0% :: 0.864] Building CXX object CMakeFiles/UnitTests.dir/Debug/cmake_pch.hxx.pch
[build] [23/179   1% :: 0.872] Building CXX object CMakeFiles/Jolt.dir/Debug/cmake_pch.hxx.pch
[build] [24/179   1% :: 0.994] Building CXX object CMakeFiles/UnitTests.dir/Debug/home/tksuoran/git/tksuoran/JoltPhysics/UnitTests/Core/FPFlushDenormalsTest.cpp.o
[build] [25/179   2% :: 1.029] Building CXX object CMakeFiles/UnitTests.dir/Debug/home/tksuoran/git/tksuoran/JoltPhysics/UnitTests/Geometry/EllipseTest.cpp.o
[build] [26/179   2% :: 1.046] Building CXX object CMakeFiles/UnitTests.dir/Debug/home/tksuoran/git/tksuoran/JoltPhysics/UnitTests/Core/InsertionSortTest.cpp.o
[build] FAILED: CMakeFiles/UnitTests.dir/Debug/home/tksuoran/git/tksuoran/JoltPhysics/UnitTests/Core/InsertionSortTest.cpp.o 
[...]

PR #262, opened 2022-10-10
Fix deprecated implicit copy constructor

tksuoran:
At least Clang 15.0.2 with `-Wextra` / `-Wdeprecated-copy` complains 'definition of implicit copy constructor is deprecated'.


jrouwe:
See comment in #261, this PR should no longer be needed.

PR #271, opened 2022-10-13
Separated body allocation / init from insertion.

sherief:
With the new CreateBodyAtID() API the save / restore problem is solved, but new issues and performance bottlenecks popped up during streaming: to guarantee deterministic IDs when streaming in a new chunk of the world, the ID allocation step needs to be serialized - but the existing API bundled memory allocation & initialization with body insertion, forcing serialization.

With this new API new objects can have their expensive allocations made in parallel on many threads, while the ID assignment can happen on a single thread in a deterministic order. This enables better use of multi-core systems during streaming while guaranteeing deterministic BodyIDs despite the unordered parallel nature of the body allocation process.

### DIFF
--- a/Jolt/Physics/Body/BodyInterface.cpp
+++ b/Jolt/Physics/Body/BodyInterface.cpp
@@ -25,6 +25,16 @@ Body *BodyInterface::CreateBodyWithID(const BodyID &inBodyID, const BodyCreation
 	return mBodyManager->CreateBodyWithID(inBodyID, inSettings);
 }
 
+Body *BodyInterface::AllocateBody(const BodyCreationSettings &inSettings)
+{
+	return mBodyManager->AllocateBody(inSettings);
+}
+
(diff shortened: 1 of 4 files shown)


jrouwe:
Hello,

I've looked at the code, but I think it has a number of issues:

1. It duplicates a large piece of code in BodyManager::InsertBody
2. I think BodyInterface::InsertBody is easily confused with BodyInterface::AddBody while they serve completely different purposes.
3. BodyInterface::InsertBody has a race condition when 2 threads try to insert a body under the same ID because mBodies is updated outside the mBodiesMutex lock (there was a similar issue in CreateBodyWithID)
4. It was possible to assign one body multiple body IDs which would lead to corruption
5. There was no way to recover from allocating a body but not being able to assign an ID to that body

I created #272 instead which should solve these issues while allowing you to do what you want to.

sherief:
I also hated the naming of these functions tbh, I like your solution better. Thanks!

PR #272, opened 2022-10-14
Added functionality to create bodies without ID

jrouwe:


sherief:
I was just working on an update - can we also separate DestroyBodies() into RemoveBodies() + Deallocate()? I'd like to be able to call RemoveBodies() from the main thread, then do the actual memory deallocation later and / or on another thread.

PR #275, opened 2022-10-15
Added ability to remove bodies from sim without deallocating them.

sherief:
This allows a quick, serial, and deterministic removal on a single thread without having to pay the cost of memory deallocation immediately.

Deallocations can be done opportunistically and / or on background thread, to prevent frame time spikes.

### DIFF
--- a/Jolt/Physics/Body/BodyInterface.cpp
+++ b/Jolt/Physics/Body/BodyInterface.cpp
@@ -52,6 +52,16 @@ bool BodyInterface::AssignBodyID(Body *ioBody, const BodyID &inBodyID)
 	return mBodyManager->AddBodyWithCustomID(ioBody, inBodyID);
 }
 
+void BodyInterface::RemoveBodyNoDeallocate(const BodyID &inBodyID)
+{
+	mBodyManager->RemoveBodies(&inBodyID, 1);
+}
+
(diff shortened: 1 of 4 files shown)


sherief:
As always, I'm not comfortable with the current function names and I'm certain they can be improved - I'm open to suggestions.

sherief:
So far: love the performance, hate the fact that on BodyInterface the function RemoveBodyNoDeallocate() does something completely different from RemoveBody().

jrouwe:
Hello,

I'm not sure if this change is really needed. First of all it duplicates a bunch of code which I would like to avoid and secondly, DestroyBodies can already be done on a background thread (that's what we do). As long as you make sure you don't immediately reuse the same Body ID then you should be fine.

Regards,

Jorrit.

sherief:
Agree on the code duplication. The problem I was trying to solve is that I sometimes don't have enough CPU time to do the deallocations before moving on to the next tick. Let's say I want to remove 1000 objects before the next sim update and my dealloc() takes (hypothetically) 1 ms - to maintain the target frame rate I need to defer the dealloc by one or more frames. While the numbers here are exaggerated, this was written to solve an actual problem that occurred in production.

jrouwe:
Ok, I'll have a look at it (I'll come up with a better interface too). First I need to fix that virtual character.

jrouwe:
I updated the interface, tried to keep the copy pasting to a minimum and submitted the change as #291.

PR #276, opened 2022-10-16
Disabled unused argument warning when compiling with clang-cl

mihe:
When compiling with `clang-cl` and using a generator other than Visual Studio (like `"Ninja Multi-Config"`) you currently get a compile error saying:

```
clang-cl: error: argument unused during compilation: '/MP' [-Werror,-Wunused-command-line-argument]
```

Here's how to reproduce it:

```
cmake -S Build -B Build/Ninja_Clang -G "Ninja Multi-Config" -DCMAKE_CXX_COMPILER=clang-cl
cmake --build Build/Ninja_Clang --config Distribution
```

This PR adds the [`-Qunused-arguments`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-qunused-arguments) argument to disable any such warnings.

If you want I can instead move the offending arguments (`/MP`, `/GL`, `/Gm-`, `/showFilenames`) behind some if-statement, but I would then suggest to add a (non-Visual Studio) `clang-cl` job to the CI builds to prevent any future arguments from causing the same problem.


jrouwe:
Thank you!

PR #285, opened 2022-10-19
Add convex radius accessors for shape classes that lack them

MDZhB:
A very minor contribution. Needed these for custom serialization, and found that they were missing.


jrouwe:
Thanks!

PR #287, opened 2022-10-19
Add convex hull point accessor

MDZhB:
Similar to my last PR. 

Another approach would have been to add a getter for an `Array<Point>`, which would have been more symmetrical with `GetPlanes()`. However, I didn't want to expose more internal state than necessary. I also considered adding a getter for an iterable, read-only view of the points as `Vec3`s, but this would have meant adding too much code for a minor feature.

### DIFF
--- a/Jolt/Physics/Collision/Shape/ConvexHullShape.cpp
+++ b/Jolt/Physics/Collision/Shape/ConvexHullShape.cpp
@@ -1186,6 +1186,12 @@ Shape::Stats ConvexHullShape::GetStats() const
 		triangle_count);
 }
 
+Vec3 ConvexHullShape::GetPoint(size_t index) const
+{
+	JPH_ASSERT(index >= 0 && index < mPoints.size(), "Point index out of bounds");
+	return mPoints[index].mPosition;
+}
+
 void ConvexHullShape::sRegister()
 {
 	ShapeFunctions &f = ShapeFunctions::sGet(EShapeSubType::ConvexHull);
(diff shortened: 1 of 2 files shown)


jrouwe:
Thanks!

Can I ask why you're not using the SaveBinaryState functions or the ObjectStreamOut class for saving the shapes?

MDZhB:
There are several reasons.

The first is that I'm using Jolt via FFI, and I felt that to mixing my language's I/O streams with C++ streams would be awkward.

Another is that it looks to me like Jolt's built-in serialization simply reads values from a stream and assigns them to private fields. This makes sense, since the `Shape` is meant to be "cooked data" that can be deserialized very quickly, but it's not what I want. My preference is that deserialized objects are constructed normally, through public APIs, as this allows them to pass through the same validation as objects constructed explicitly in code (and I'm okay with eating the cost of that validation).

I could have achieved this with `ShapeSettings`, but hanging on to the settings object would effectively mean maintaining two copies of every shape throughout its lifetime.

Regarding your additional commits: thank you for cleaning up my code. I don't get to write much C++, so I tried to be very careful about not defacing your API. Evidently I wasn't as thorough as I could have been.

It was my understanding that `std::vector`'s subscript operator does not do bounds checking, whereas `at()` does. My intent in using the assert and the subscript operator was that application code known to be well-behaved could opt out of the bounds check. Please let me know where I went wrong.

jrouwe:
Ok, thanks for the explanation!

It seems that we're both right about std::vector, I've just tested this piece of code:

```	
vector<int> x;
x.resize(10);
printf("%d\n", x[11]);
```

And in a Debug build on Windows it triggers: `Expression: vector subscript out of range`. In Release mode it doesn't complain.
On Linux it happily reads out of bounds.

Replacing x[11] with at(11) throws std::out_of_range on both platforms.

So an assert would have been the safe thing to do after all. Learned something new today :)

MDZhB:
That explains it -- I build with MinGW on Windows, so I use the GNU implementation on both platforms.

Thank you for your quick responses.

PR #290, opened 2022-10-21
AVX512 tweaks

jrouwe:


jrouwe:
@Wunkolo I see that you have been making some further AVX512 changes. Do you consider these mergeable? The reason I'm asking is that probably next week I'm going to work on this class so I'd like to get any pending changes in before that.

Wunkolo:
Yes this is ready to merge. I was holding off for a bit as I was experimenting with introducing a "bool-vector"(Similar to `bvec` in glsl) to store true/false and comparison results to try and alias into k-mask registers in the case that AVX512 is enabled to see if there were gains there, but otherwise this is all good to go 👌.

jrouwe:
@Wunkolo I noticed that if I merge this it will be submitted under my name because I created the PR. Can you perhaps initiate a pull request? (I don't want to steal your credits)

PR #305, opened 2022-10-27
Added CMake multi-config support for GCC

mihe:
When compiling with `g++` and using a multi-config generator like `"Ninja Multi-Config"` you currently run into CMake errors such as:

```
ninja: error: loading 'build-Distribution.ninja': No such file or directory
```

Here's how to reproduce it (you'll need the [`ninja-build`](https://packages.ubuntu.com/jammy/ninja-build) package):

```
cmake -S Build -B Build/Ninja_GCC -G "Ninja Multi-Config" -DCMAKE_CXX_COMPILER=g++
cmake --build Build/Ninja_GCC --config Distribution
```

This seems to stem from `CMAKE_CONFIGURATION_TYPES` not being set for anything but MSVC and Clang currently.

This PR sets that to be `Debug;Release;Distribution` for GCC as well.

GCC does seem to have support for ASan and UBSan, but not for code coverage, so I kept it simple and left all of them out.


jrouwe:
Thanks!

PR #310, opened 2022-10-29
Added missing va_end to va_list usages

mihe:
As mentioned on [cppreference.com](https://en.cppreference.com/w/c/io/vfprintf#Notes) regarding `vsnprintf`:

> These functions do not invoke `va_end`, and it must be done by the caller.

It seems to just null out the `va_list` pointer on MSVC, so no leaks or anything, but I haven't checked other platforms.


jrouwe:
Thanks!

PR #356, opened 2022-12-12
Added ScaledShapeSettings constructor which takes concrete shape

mihe:
This one seems to have been accidentally left out, as the other decorator shapes have a similar constructor already.


jrouwe:
Thanks!

PR #357, opened 2022-12-12
Added CMake option for link-time optimizations

mihe:
I'm currently unable to disable link-time code generation (`/LTCG`) on MSVC builds due to Jolt unconditionally using `/GL` for the `Release` and `Distribution` configurations. Microsoft's linker forcefully switches over to `/LTCG` mode whenever it finds an object file or static library with `/GL` enabled.

I figured instead of hiding the current use of `/GL` behind an `option` I would solve this using CMake's built-in support for adding these types of flags, which CMake calls [`INTERPROCEDURAL_OPTIMIZATION`](https://cmake.org/cmake/help/v3.16/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html#variable:CMAKE_INTERPROCEDURAL_OPTIMIZATION).

This has the added benefit of also adding the appropriate flags for GCC and Clang, where the equivalent of `/GL` is `-flto` and `-flto=thin` respectively.

I also made use of CMake's suggested [`check_ipo_supported`](https://cmake.org/cmake/help/v3.16/module/CheckIPOSupported.html) helper function, which does a `try_compile` with IPO enabled to see if the current platform/compiler supports it. Sadly this does not seem to catch the internal compiler error that happens when using IPO on Windows ARM/ARM64, so I'm not sure how much value it really provides.

Along with removing the explicit `/LTCG` I also removed the explicit `/INCREMENTAL:NO`, `/OPT:ICF` and `/OPT:REF`, all of which seem to be the default values of those flags when `/DEBUG` isn't specified.

It might also be worth mentioning that when using a Visual Studio generator, CMake simply emits `<WholeProgramOptimization>` in the project files, which Visual Studio then translates into `/GL` and `/LTCG:INCREMENTAL`, meaning you don't get the regular `/LTCG`. This only affects Jolt's executables and shared libraries though, which were already using the default `/LTCG:INCREMENTAL` previously.

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -21,6 +21,9 @@ option(CROSS_PLATFORM_DETERMINISTIC "Cross platform deterministic" OFF)
 # When turning this option on, the library will be compiled for ARM (aarch64-linux-gnu), requires compiling with clang
 option(CROSS_COMPILE_ARM "Cross compile to aarch64-linux-gnu" OFF)
 
+# When turning this option on, the library will be compiled with interprocedural optimizations enabled, also known as link-time optimizations or link-time code generation
+option(INTERPROCEDURAL_OPTIMIZATION "Enable interprocedural optimizations" ON)
+
 # Select X86 processor features to use (if everything is off it will be SSE2 compatible)
 option(USE_SSE4_1 "Enable SSE4.1" ON)
 option(USE_SSE4_2 "Enable SSE4.2" ON)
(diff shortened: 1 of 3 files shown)


jrouwe:
I'm afraid that this now makes clang smart enough to realize that TestFltMin in FPFlushDenormalsTest.cpp is not referenced anywhere else so all the calculations in that file are replaced with constants, meaning it doesn't respect the flush denormals FPU flag anymore. I have a workaround for that though.

jrouwe:
One more issue left.

I think this compiler error is wrong (it's readying 1 byte from a 1 byte atomic, not writing 1 byte to a 0 byte value):

```
In member function 'load',
    inlined from 'IsSensor' at D:/a/JoltPhysics/JoltPhysics/Jolt/Physics/Body/Body.h:68:66,
    inlined from 'AddHit' at D:\a\JoltPhysics\JoltPhysics\Jolt\Physics\Vehicle\VehicleCollisionTester.cpp:131:23:
D:/a/_temp/msys64/mingw64/include/c++/12.2.0/bits/atomic_base.h:488:31: error: '__atomic_load_1' writing 1 byte into a region of size 0 overflows the destination [-Werror=stringop-overflow=]
  488 |         return __atomic_load_n(&_M_i, int(__m));
```

Suggestions? (probably best to add a AND NOT MINGW on line 240).

jrouwe:
The joy of supporting 10000 compiler versions :)

jrouwe:
Thanks for the investigations!

PR #360, opened 2022-12-15
Added ShapeFilter param to various CharacterVirtual functions

DocAce:


jrouwe:
Thanks, that's indeed something I forgot to expose.

PR #432, opened 2023-02-09
Allowed sensors to collide with one another

mihe:
I figured I could ask this question in the form of a PR. :)

I noticed that sensors are unable to detect other sensors overlapping with them. I realize this is probably an unusual thing to want to have, but it is none the less something that Godot offers, so I must try to implement it.

I dug through the code a bit and found the lines that I've removed in this PR. Removing them seems to unlock the ability to receive sensor-to-sensor contact reports.

Is there a particular reason why this should be there, considering that the existing filtering mechanisms, like `ObjectLayerPairFilter`, have the opportunity to filter out those collisions already?

Feel free to push back on this if it seems too niche of a use-case. It's a small enough change that I'd be happy to maintain it on my own fork, assuming it even makes sense to begin with.

### DIFF
--- a/Jolt/Physics/Body/Body.inl
+++ b/Jolt/Physics/Body/Body.inl
@@ -31,11 +31,9 @@ inline bool Body::sFindCollidingPairsCanCollide(const Body &inBody1, const Body 
 	// One of these conditions must be true
 	// - One of the bodies must be dynamic to collide
 	// - A kinematic object can collide with a sensor
-	bool body1_sensor = inBody1.IsSensor();
-	bool body2_sensor = inBody2.IsSensor();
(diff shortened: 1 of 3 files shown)


jrouwe:
Hello,

So you're not the first person to mention this: https://github.com/jrouwe/JoltPhysics/discussions/282

Do you know what this is used for in godot?

jrouwe:
I did some further testing and I agree with you that it is really easy to disable the collision between sensors if you don't want them to interact. Thanks for the change!

jrouwe:
I ensured that the SensorTest sample and the sensor unit tests all use a separate layer for the sensors, so if people copy paste from there it should be ok. I think the HelloWorld example should be a minimal example, so I don't want to add sensors there.

PR #461, opened 2023-03-10
Update JobSystem.h documentation

BobbyAnguelov:
Just one additional comment to the job system documentation that is obvious in hindsight but wasn't immediately obvious to me when I integrated it.


jrouwe:
Thanks!

PR #495, opened 2023-04-06
Add a way to get a convex hull edges by exposing its faces

SirLynix:
I was trying to get a convex hull edges but I could only retrieve its points (or triangles) where I'm only interested in its edges.

This is a simple PR exposing a convex hull face vertices, which allows to easily get its edges.


jrouwe:
Hello,

I don't think it can work like this. GetFaceVertices returns the first vertex and number of vertices in mVertexIdx but this array is not exposed so I don't think you can determine anything from this (my guess is that you were assuming these indexes are for GetPoints but there's an extra layer of indirection).

I'll patch it up.

SirLynix:
Indeed I though those were points indices.

Thanks a lot for fixing that and merging this PR!

Issue #494, opened 2023-04-06
ARMv7 and package manager support (xrepo)


SirLynix:
Hi.

I added Jolt Physics to xmake-repo, a package manager for C++ (see https://github.com/xmake-io/xmake-repo/pull/1685).

[XMake](https://xmake.io) is a build system (like CMake) which is able to use other build systems, for example the [Bullet script](https://github.com/xmake-io/xmake-repo/blob/master/packages/b/bullet3/xmake.lua#L25) directly triggers cmake and then xmake retrieves the built includes/binaries.

I tried to integrate Jolt the same way but this was a bit complicated for multiple reasons, one of them is that Jolt CMakeLists forces MSVC runtime library.

```cmake
	# Set runtime library
	if (USE_STATIC_MSVC_RUNTIME_LIBRARY)
		set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
	else()
		set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
	endif()
```

XMake allows one to build a debug application with release packages (which is useful to debug app without having all dependencies without optimization), but since runtime library "must" match across binaries it allows to compile dependencies in release with a debug runtime.

For example, when I'm debugging my engine, I'm using a debug runtime (MDd) but my dependencies are still compiled in release (to keep optimization active on them, Assimp is really slow in debug mode) with the MDd runtime (since Jolt is statically linked, it has to match).

I had other issues with platform detection, so instead of patching the CMakeLists I decided to rewrite the build script for xmake: https://github.com/xmake-io/xmake-repo/blob/master/packages/j/joltphysics/port/xmake.lua

This is only to give a bit of context and to tell you that Jolt is available on another package manager/build system.

Now, that build script works well for Windows/Linux/macOS/iOS/emscripten and Android (ARMv8-a only) but it [fails on some platforms](https://github.com/SirLynix/xmake-repo/pull/6):

- [Windows ARM64 using Visual Studio 2019](https://github.com/SirLynix/xmake-repo/actions/runs/4448172888/jobs/7810631455?pr=6#step:5:44) (some intrinsics were not available for ARM64 before VS2022, mostly don't care).
- [Android ARMv7](https://github.com/SirLynix/xmake-repo/actions/runs/4448172922/jobs/7810630422?pr=6#step:5:163)
[...]

PR #496, opened 2023-04-08
Removed setting of MSVC runtime library when using DLL runtime

mihe:
This removes the redundant setting of `CMAKE_MSVC_RUNTIME_LIBRARY` when using the dynamically linked MSVC runtime, which allows the consuming project to set it instead, like when mixing debug and release runtimes.

Related to #494.


jrouwe:
Thanks, that looks like a good idea!

PR #513, opened 2023-04-18
Support for building Jolt as a shared library on Windows

jankrassnigg:
Adds a CMake option "COMPILE_AS_SHARED_LIBRARY" to switch the Jolt library to be compiled as a shared library.

This sets two compiler flags
* JPH_SHARED_LIBRARY is a public define, that instructs all users of the library to import symbols
* JPH_BUILD_SHARED_LIBRARY is a private define, that instructs the Jolt library to export its symbols

These can easily be set as well by custom build tools.

All classes, global variables and functions have been marked with JPH_EXPORT, such that the important symbols get exported. I think I haven't missed anything important, it seems to all be working.

static thread_local variables can't be DLL exported (IMO this is a limitation of the compiler), so I've wrapped a few of these in static functions.

The RTTI macros declare a global function "GetRTTIOfType" which also has to be exported. Therefore I had to extend those declaration macros to take the export macro as a parameter. This is a bit unfortunate, I had hoped that such a duplication wouldn't be necessary.

All other libraries continue to use static linking (or are binaries anyway), so where they use those RTTI macros, they pass in JPH_NO_EXPORT as a no-op.


Jolt/Core/Profiler.inl
```diff
@@ -25,15 +25,15 @@ ProfileThread::~ProfileThread()
 
 ProfileMeasurement::ProfileMeasurement(const char *inName, uint32 inColor)	
 {
-	if (ProfileThread::sInstance == nullptr)
+	if (ProfileThread::Instance() == nullptr)
```
jrouwe, line 28: These function calls are problematic as it is part of the profiling system, so a function call overhead skews the measuring results.

Jolt/Physics/Body/BodyManager.cpp
```diff
@@ -20,8 +20,18 @@
 JPH_NAMESPACE_BEGIN
 
 #ifdef JPH_ENABLE_ASSERTS
-	thread_local bool BodyManager::sOverrideAllowActivation = false;
-	thread_local bool BodyManager::sOverrideAllowDeactivation = false;
+	static thread_local bool sOverrideAllowActivation = false;
+	static thread_local bool sOverrideAllowDeactivation = false;
+
+	bool& BodyManager::OverrideAllowActivation()
```
jrouwe, line 26: Please make these functions sSetOverrideAllowActivation/sGetOverrideAllowActivation (same for deactivation) to be more consistent with the rest of the code.

Jolt/Math/UVec4.h
```diff
@@ -212,7 +212,7 @@ class [[nodiscard]] alignas(JPH_VECTOR_ALIGNMENT) UVec4
 	};
 
 private:
-	static const UVec4			sFourMinusXShuffle[];
+	JPH_EXPORT static const UVec4	sFourMinusXShuffle[];
```
jrouwe, line 215: This means that a call to ShiftComponents4Minus will have to fetch the pointer to sFourMinusXShuffle and then dereference it to fetch the actual address of the sFourMinusXShuffle table, whereas before it would be able to load the constant address of the sFourMinusXShuffle table directly. This is problematic as the intrinsics are usually used in tight inner loops.

Build/CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ option(CROSS_PLATFORM_DETERMINISTIC "Cross platform deterministic" OFF)
 # When turning this option on, the library will be compiled for ARM (aarch64-linux-gnu), requires compiling with clang
 option(CROSS_COMPILE_ARM "Cross compile to aarch64-linux-gnu" OFF)
 
+# When turning this option on, Jolt will be compiled as a shared library and public symbols will be exported.
+option(COMPILE_AS_SHARED_LIBRARY "Compile Jolt as a shared library" OFF)
```
jrouwe, line 18: The 'Distribution' build currently gives a compile error if you turn this on (because I'm doing some tricks with DebugRenderer.h).

Jolt/Core/Core.h
```diff
@@ -4,6 +4,22 @@
 
 #pragma once
 
+// If this define is set, Jolt is compiled as a shared library
+#ifdef JPH_SHARED_LIBRARY
+	#ifdef JPH_BUILD_SHARED_LIBRARY
+		// While building the shared library, we must export these symbols
+		#define JPH_EXPORT __declspec(dllexport)
```
jrouwe, line 11: How compatible is this with Linux? Does everything need to change if someone asks for a shared library on Linux?

jrouwe:
Another thing that I noticed: The debug symbols are missing from the Release (and probably also the Distribution) version of the dll making it really hard to debug them.

Jolt/Math/UVec4.h
jankrassnigg: Since it's only used in UVec4::ShiftComponents4Minus, would you agree to just put the table into the inline function instead?

Jolt/Core/Core.h
jankrassnigg: Looking at our code, in Linux you would use "[[gnu::visibility("default")]]" rather than __declspec(dllimport)/__declspec(dllexport). The rest stays the same.

Jolt/Math/UVec4.h
jankrassnigg: Regarding the pointer fetch: are you certain about that or is it an assumption? Currently when I leave out the JPH_EXPORT, only the UnitTests project fails to link, the others (e.g. the Samples) work fine. That's because the UnitTests call those inline functions and therefore need to link to the table.
My point is, that pointer fetch should only be necessary for external code. Everything internal to the Jolt library should still be able to load the data directly (though I haven't looked at any generated assembly). Since all the performance critical code is in Jolt, it shouldn't make a difference and the export is only necessary so that the unit tests still compile (which IMO aren't that performance critical).

Build/CMakeLists.txt
jankrassnigg: Yeah, I managed to make those tricks work, but it's not pretty.

Jolt/Core/Profiler.inl
jankrassnigg: I changed it so that in static builds it's a direct access (well, via an inlined function) and in shared builds it isn't. There is no way to make this work in a shared build without the indirection (as far as I can tell).

Jolt/Core/Core.h
mihe: Don't take my word for this, but I believe you should also (ideally) pair that up with the compiler flag `-fvisibility=hidden` on Linux, to stop it from bundling any symbols that aren't explicitly exported. See [this link](https://gcc.gnu.org/wiki/Visibility) for more details, as well examples of cross-platform export macros.

Jolt/Math/UVec4.h
jrouwe: I looked at the disassembly yesterday, and it is indeed only an issue outside of the Jolt dll. Inside the Jolt dll it uses a ```lea register, constant address``` while outside it uses a ```mov register, address to pointer```.

I tried putting it inside UVec4::ShiftComponents4Minus as a local static but that gives horrible static initialization code using a thread local:

```
00007FF69B811A6B  mov         ecx,4  
00007FF69B811A70  mov         rax,qword ptr gs:[58h]  
00007FF69B811A79  mov         r8,qword ptr [rax]  
00007FF69B811A7C  mov         eax,dword ptr [rcx+r8]  
00007FF69B811A80  cmp         dword ptr [$TSS0 (07FF69BA49E40h)],eax  
00007FF69B811A86  jle         07FF69B811AD8h  
00007FF69B811A88  lea         rcx,[$TSS0 (07FF69BA49E40h)]  
00007FF69B811A8F  call        _Init_thread_header (07FF69B956EA8h)  
00007FF69B811A94  cmp         dword ptr [$TSS0 (07FF69BA49E40h)],0FFFFFFFFh  
00007FF69B811A9B  jne         07FF69B811AD8h  
00007FF69B811A9D  vpcmpeqd    xmm0,xmm0,xmm0  
00007FF69B811AA1  vmovdqu     xmmword ptr [sFourMinusXShuffle (07FF69BA49E50h)],xmm0  
00007FF69B811AA9  vmovaps     ymm1,ymmword ptr [__ymm@ffffffffffffffff0f0e0d0c0b0a0908ffffffffffffffffffffffff0f0e0d0c (07FF69B9FC660h)]  
00007FF69B811AB1  vmovups     ymmword ptr [sFourMinusXShuffle+10h (07FF69BA49E60h)],ymm1  
00007FF69B811AB9  vmovaps     ymm1,ymmword ptr [__ymm@0f0e0d0c0b0a09080706050403020100ffffffff0f0e0d0c0b0a090807060504 (07FF69B9FA8C0h)]  
00007FF69B811AC1  vmovups     ymmword ptr [sFourMinusXShuffle+30h (07FF69BA49E80h)],ymm1  
00007FF69B811AC9  lea         rcx,[$TSS0 (07FF69BA49E40h)]  
00007FF69B811AD0  vzeroupper  
00007FF69B811AD3  call        _Init_thread_footer (07FF69B956E48h)  
00007FF69B811AD8  vpshufb     xmm0,xmm6,xmmword ptr [sFourMinusXShuffle+20h (07FF69BA49E70h)]  
```

Using ```static constexpr UVec4 sFourMinusXShuffle[] = ...``` inside the function declaration doesn't work either since the intrinsics are not constexpr, so I went with ```static constexpr uint32 sFourMinusXShuffle[][]``` aligned to 16 bytes and that seems to work.

Jolt/Core/Core.h
jrouwe: I tried ```[[gnu::visibility("default")]]``` yesterday and then clang started generating errors that this attribute is invalid on a class but today it works without errors. @mihe I followed the steps you list and now it produces a .so on Linux, thanks!

jrouwe:
So I think there's 2 issues left:

1. In the DLL version in the Release and Distribution configs there is 1 unit test that fails for some reason:
```
CharacterVirtualTests.cpp(492):
TEST SUITE: CharacterVirtualTests
TEST CASE:  TestContactPointLimit

CharacterVirtualTests.cpp(560): ERROR: CHECK( character.mCharacter->GetActiveContacts().size() < character.mCharacter->GetMaxNumHits() ) is NOT correct!
  values: CHECK( 256 <  256 )
```

This does not happen on the .so version or on any of the statically linked versions.

2. The Release and Distribution version of Jolt.dll don't have symbols. The ```/DEBUG:FASTLINK``` linker option is missing. Not sure what the CMake way to get it there is (adding it to CMAKE_SHARED_LINKER_FLAGS_RELEASE?).

jrouwe:
> I would assume that this is due to different floating point precision somewhere. Unfortunately lots of things can affect that.

Turned out to be a stupid mistake in the unit test itself :)

jrouwe:
I think it's all good now. Thanks for your help!

jankrassnigg:
That option is new to me. But from the documentation this sounds like an all or nothing option, which would be unacceptable when integrating into other projects. For example, I have integrated Jolt into another engine and I need Jolt to be a DLL, but I don't want to set a CMake option that globally affects the build of my other 100 libraries only because of this. So an extra option just for Jolt prevents issues when integrating somewhere else.

SirLynix:
I'm not familiar with FetchContent but doesn't cmake provide a way to set option per dependency? Many libraries use BUILD_SHARED_LIBS and I find it weird if it has such a restriction.

You should be able to set BUILD_SHARED_LIBS for a single dependency without forcing it on your project.

SirLynix:
How are you integrating Jolt in your project?

If using add_subdirectory, you could `set` BUILD_SHARED_LIBS and unset it after without issue.
```
set(BUILD_SHARED_LIBS ON)
add_subdirectory(Jolt)
set(BUILD_SHARED_LIBS OFF)
```

I don't think `COMPILE_AS_SHARED_LIBRARY` helps here because, what if another CMake projects uses this config name? You'll have the same issue.

PR #517, opened 2023-04-27
Add missing standard include

SirLynix:
Jolt uses `std::uintX_t` types, which are defined in the [cstdint](https://en.cppreference.com/w/cpp/header/cstdint) header without including it.

This works on most platforms because this header is included by other standard includes (but we have no guarantee on that), and Jolt compilation recently failed on Linux Fedora because of a compiler update. (see https://github.com/SirLynix/xmake-repo/actions/runs/4817052207/jobs/8577304087 )

This commits fixes it (tested [here](https://github.com/SirLynix/xmake-repo/actions/runs/4817136733/jobs/8577484264) by [applying this commit as a patch](https://github.com/SirLynix/xmake-repo/pull/9/files#diff-df1ba7aad51e5146e071f1b2d7d74c0fc9f2a65cd972fa72d8c25277b976cd16R11)).

On a side note, I'm wondering about these includes:
```cpp
#include <limits.h>
#include <float.h>
#include <string.h>
```

is there a reason Jolt doesn't include the corresponding C++ headers?
```cpp
#include <climits>
#include <cfloat>
#include <cstring>
```

### DIFF
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -296,6 +296,7 @@ JPH_SUPPRESS_WARNINGS_STD_BEGIN
 #include <sstream>
 #include <functional>
 #include <algorithm>
+#include <cstdint>
 JPH_SUPPRESS_WARNINGS_STD_END
 #include <limits.h>
 #include <float.h>


jrouwe:
Thanks!

jrouwe:
About limits.h -> climits etc: I'm old school, I have been using them forever and didn't know that they had replacements (also C++ headers are often bulkier and slower to compile, so I'm not sure if it's worth it to replace them).

SirLynix:
C++ has all C headers in the c form (`<string.h>` => `<cstring>` for example), they shouldn't be heavier as all they should do is wrap C headers in std namespace (and add some overloads, like float overload for `pow`, or make some pointers const), while keeping C functions global.

It's commonly seen as a good practice to use them (see https://stackoverflow.com/questions/8749003/including-standard-header-files-string-h-or-cstring-or-both)

PR #518, opened 2023-04-27
Replace COMPILE_AS_SHARED_LIBRARY by the standard CMake BUILD_SHARED_LIBS

SirLynix:
As discussed in https://github.com/jrouwe/JoltPhysics/pull/513

Instead of introducing a new option name, reuse `BUILD_SHARED_LIBS` [from CMake](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html), which is pretty common.

I also made it generate debug symbols only if `GENERATE_DEBUG_SYMBOLS` was on in shared mode.


Jolt/Jolt.cmake
```diff
@@ -429,34 +429,33 @@ endif()
...
+			set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} -g")
+		endif()
 
-	# Set linker flags for other build types to be the same as release
-	set(CMAKE_SHARED_LINKER_FLAGS_RELEASEASAN "${CMAKE_SHARED_LINKER_FLAGS_RELEASE}")
-	set(CMAKE_SHARED_LINKER_FLAGS_RELEASEUBSAN "${CMAKE_SHARED_LINKER_FLAGS_RELEASE}")
-	set(CMAKE_SHARED_LINKER_FLAGS_RELEASECOVERAGE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE}")
-	set(CMAKE_SHARED_LINKER_FLAGS_DISTRIBUTION "${CMAKE_SHARED_LINKER_FLAGS_RELEASE}")
+		# Set linker flags for other build types to be the same as release
+		set(CMAKE_SHARED_LINKER_FLAGS_RELEASEASAN "${CMAKE_SHARED_LINKER_FLAGS_RELEASE}")
```
jrouwe, line 448: I think these set statements need to be outside if GENERATE_DEBUG_SYMBOLS because otherwise the ASAN/UBSAN etc. builds will not build as the variables are undefined.

SirLynix: I'm not a CMake expert but as far as I understand it, they will just be empty and ignored, compilation should be fine.

jrouwe: I had to add them explicitly last time because it wouldn't build :)

jankrassnigg:
> @jankrassnigg I did not see consensus in #513, since you created the original change, are you ok with this change?

Sorry, too many things happening in my life atm to keep on top of everything.
As I said the BUILD_SHARED_LIBS option has so far been unknown to me, so I can't say whether it has any downsides, but I trust that it just works.

So yeah, I'm fine with this change.

jrouwe:
Ok, thanks!

This is a change I'd like to test myself as well but I won't have access to a setup where I can test changes for the next 1.5 weeks, so I'll leave this PR until then.

jrouwe:
I've tested the changes, thanks for making them!

PR #550, opened 2023-05-27
Allow negative friction on a body

jrouwe:


jrouwe:
See discussion at godot-jolt/godot-jolt#396

PR #581, opened 2023-06-17
Sensors now can detect static objects

jrouwe:


mihe:
I seem to trigger this assertion in `NonReductionCollideShapeCollector` when running with `mSensorDetectsStatic` enabled and `mUseManifoldReduction` disabled, right at the point where the sensor starts overlapping with the static body:

https://github.com/jrouwe/JoltPhysics/blob/65ca648ae8142ef56d42803da5cce3fe4cfdc000/Jolt/Physics/PhysicsSystem.cpp#L1159

In my case `mBody1` is the static body and `mBody2` is the kinematic sensor.

I'd be happy to create a repro if you need it.

jrouwe:
Fixed!

mihe:
I found another strange interaction.

If you add this to the bottom of `HeightFieldShapeTest::Initialize`:

```cpp
BodyCreationSettings sensor_settings(new BoxShape(Vec3::sReplicate(5.0f)), RVec3(0, 0, 0), Quat::sIdentity(), EMotionType::Kinematic, Layers::MOVING);
sensor_settings.mIsSensor = true;
sensor_settings.mSensorDetectsStatic = true;
mBodyInterface->CreateAndAddBody(sensor_settings, EActivation::Activate);
```

You'll get an assert about "Unsupported shape pair", where `inShape1` is the `HeightFieldShape` and `inShape2` is the sensor's `BoxShape`.

jrouwe:
Fixed. I have to say though that this can be very expensive. You will get a contact callback for every triangle in the height field, so if you have a large sensor colliding with the heightfield then that's a lot of callback (and a lot of memory to store all those contacts).

mihe:
Thank you for the fix.

Yeah, I turned on the contact visualization in Godot just now, and the overlap between the area and the (fairly low-res) height map lights up like a christmas tree. That's definitely more than a little disconcerting.

I don't really have any good way of prominently displaying warnings about this kind of stuff from the extension side of things either, unless I start abusing project settings for some sort of acknowledgement from the user.

I'll try my best to make it clear in the documentation.

jrouwe:
How does that work with godot physics: Does it also light up? (or is this a godot-jolt specific visualization?)

mihe:
> is this a godot-jolt specific visualization?

No, it's a part of Godot's physics server interface. You enable it by toggling "Visible Collision Shapes" in the "Debug" menu in the editor.

> Does it also light up?

Areas in Godot Physics don't seem to utilize contacts at all, and therefore don't show anything in the contact visualization. Instead they seem to have [bespoke body pairs](https://github.com/godotengine/godot/blob/master/servers/physics_3d/godot_space_3d.cpp#L1040-L1054) specifically for areas, which I guess skirts around the whole issue of needing contacts in the first place.

PR #591, opened 2023-06-23
fix assert in GJKClosestPoint::GetClosestPoints()

NicSavichev:
for case of non-perfect accuracy in 32 bits


jrouwe:
Hello,

ioV.LengthSq() must be exactly equal to v_len_sq because v_len_sq is initialized everywhere to ioV.LengthSq(), this is just validating that they are kept in sync. If they are not equal then your compiler may have taken some shortcuts that it shouldn't have. Maybe you're compiling with -ffast-math? In any case v_len_sq becomes much much smaller than 1.0e-6 so changing the code like this effectively turns the assert off.

W.r.t. to the unused parameters: Can you make this a separate PR and use [[maybe_unused]] instead of commenting out the parameter names?

Regards,

Jorrit.

NicSavichev:
Yes, -fp:precise (instead of default -fp:fast) for VC helped

PR #594, opened 2023-06-23
fix 'unused parameter' compiler warnings (got under GCC 11, linux)

NicSavichev:
now using C++17 [[maybe_unused]]


jrouwe:
Thanks for submitting this!

PR #602, opened 2023-07-02
Ability to specify which DOFs a body has

jrouwe:


jrouwe:
@mihe I think this one is useful for you. It allows you to restrict the degrees of freedom without having to create a 6DOF constraint.

mihe:
Sweet! I've been keeping an eye on it. I gave it a try shortly after you created the PR, but ran into some NaNs. I'll see if it works better now.

mihe:
Considering this assert triggering I'm assuming that omitting all translation DOFs (but leaving the rotation DOFs) is out of the question?

https://github.com/jrouwe/JoltPhysics/blob/2dd3a033a41e422eb470484029324cc9bbaf0825/Jolt/Physics/Body/MotionProperties.cpp#L120

Also, I guess that the trick with zeroing out the inverse inertia (mentioned in godot-jolt/godot-jolt#443) for the `lock_rotation` property in Godot can now be replaced with just omitting the rotation DOFs?

jrouwe:
It should be possible to have only rotation DOFs, that would mean that mInvInertiaDiagonal is non zero and the assert would not trigger.

You indeed don't need to do any tricks with inertia anymore, that is now moved inside MotionProperties::SetMassProperties.

And you're right that this solution is not quite the same as godot (I already realized this when making it). It should not be hard to fix this though. One way would be to place the shape in a RotateTranslatedShape with the initial orientation as requested by godot and then setting the rotation of the Joly body to identity. The same effect can be achieved by calling Shape::GetMassProperties, modify the inertia and then calling MotionProperties::SetMassProperties and modifying mInertiaRotation afterwards. I need to play around with the math a bit to give you the exact formula for that.

jrouwe:
Updating AllowedDOFsTest with this code seems to do the trick (see code marked ```THIS IS THE CODE YOU NEED```):

```
void AllowedDOFsTest::Initialize()
{
	// Floor
	CreateFloor();

	Vec3 box_size(0.5f, 1.0f, 2.0f);
	RefConst<Shape> box_shape = new BoxShape(box_size);

	for (int allowed_dofs = 1; allowed_dofs <= 0b111111; ++allowed_dofs)
	{
		float x = -35.0f + 10.0f * (allowed_dofs & 0b111);
		float z = -35.0f + 10.0f * ((allowed_dofs >> 3) & 0b111);

		// Create body 
		BodyCreationSettings bcs(box_shape, RVec3(x, 10, z), Quat::sRotation(Vec3::sReplicate(sqrt(1.0f / 3.0f)), DegreesToRadians(45.0f)), EMotionType::Dynamic, Layers::MOVING);
		Body &body = *mBodyInterface->CreateBody(bcs);
		mBodyInterface->AddBody(body.GetID(), EActivation::Activate);
		mBodies.push_back(body.GetID());

		// Update the rotation DOFs to operate in world space rather than local space
		// <<< THIS IS THE CODE YOU NEED >>>
		MotionProperties *mp = body.GetMotionProperties();
		MassProperties mass_properties = body.GetShape()->GetMassProperties();
		Mat44 rotation = Mat44::sRotation(body.GetRotation());
		mass_properties.mInertia = rotation.Multiply3x3(mass_properties.mInertia.Multiply3x3RightTransposed(rotation)); // Take the inertia to world space
		mp->SetMassProperties((EAllowedDOFs)allowed_dofs, mass_properties); // Set elements to zero
		mp->SetInverseInertia(mp->GetInverseInertiaDiagonal(), body.GetRotation().Conjugated() * mp->GetInertiaRotation()); // Now take it back to local space
		// <<< END OF THE CODE YOU NEED >>>

		// Create a constraint
		DistanceConstraintSettings dcs;
		dcs.mPoint1 = bcs.mPosition + Vec3(5, 5, 5);
		dcs.mPoint2 = bcs.mPosition + box_size;
		dcs.mMinDistance = 0.0f;
		dcs.mMaxDistance = sqrt(3.0f) * 5.0f + 1.0f;
		mPhysicsSystem->AddConstraint(mBodyInterface->CreateConstraint(&dcs, BodyID(), body.GetID()));
	}
}
```

mihe:
> this code seems to do the trick

Thank you! I'll give that a try shortly.

> It should be possible to have only rotation DOFs, that would mean that mInvInertiaDiagonal is non zero and the assert would not trigger.

Yes, you're right. It appears I was a bit ham-fisted in my testing and applied the DOFs to all the different shapes I support, and apparently it's my custom ray shape that's causing trouble and ends up with a zero `mInvInertiaDiagonal` because I'd [left out](https://github.com/godot-jolt/godot-jolt/blob/5ee6fafe9248b846814be9fc9b48a32d32289032/src/shapes/jolt_custom_ray_shape.hpp#L61) the `GetMassProperties` implementation for it, causing trouble when it fails `DecomposePrincipalMomentsOfInertia` in `SetMassProperties`.

Apparently this has been silently failing in `DecomposePrincipalMomentsOfInertia` since I first implemented it. It's not really a shape you'd use in dynamic bodies, and with it having no thickness it's not obvious to me what the correct values would be, so I'm not sure how much it really matters that it takes the "inertia tensor of sphere with radius 1" route. Anyway, implementing it using something like `SetMassAndInertiaOfSolidBox` seems to resolve the issue.

jrouwe:
> with it having no thickness it's not obvious to me what the correct values would be

Something that has no thickness has zero volume so should have zero mass. Also it makes the inertia tensor singular. Since I store the inverse of mass and inverse of inertia this cannot be supported (and it's not a sensible configuration anyway).

jrouwe:
An arbitrary mass / inertia tensor should be fine for these cases I think. For kinematic bodies the mass is infinite by default, so it doesn't matter what you specify.

mihe:
From my limited testing it seems that the back-and-forth transforming of the inertia works, which is awesome, but I am however noticing now that any ongoing linear/angular velocity (i.e. after the body has been created and is moving around) doesn't actually get set to 0 when you then restrict that particular DOF.

I was thinking this should maybe be done in `SetMassProperties`, but seeing as how these rotation DOFs are in local-space I assume you would also need to reset the respective velocity in local-space, which I guess would throw a wrench into my setup with this new transforming-of-inertia trick.

I'll do it on my end for now.

jrouwe:
The problem is that MotionProperties class has no idea of the body to which it is attached, so it doesn't know how to transform world space angular velocity to local space. I'm thinking about changing SetMassProperties to also have a body rotation parameter and perhaps an option to tell it to restrict inertia in world space.

Until that time it is indeed best to reset the velocity on your side.

PR #603, opened 2023-07-05
Added more bitwise operators and an invalid value to EAllowedDOFs

mihe:
This adds XOR and NOT bitwise operators to `EAllowedDOFs`, as well as assignment operators for all the bitwise operations. It also changes the operators to be `constexpr` (which implies `inline`) to allow them to be used in `constexpr` expressions.

I also added a `None` value to the enum, to make certain bitwise operations cleaner. I'm not sure how you feel about this one, since it does technically make it a little bit easier to shoot yourself in the foot by locking all the translation axes, but seeing as how I managed to shoot myself in the foot in this exact manner from just thinking that you might have left this out by mistake then maybe it's better to have it be there with an appropriate comment instead.

### DIFF
--- a/Jolt/Physics/Body/AllowedDOFs.h
+++ b/Jolt/Physics/Body/AllowedDOFs.h
@@ -45,21 +45,21 @@ constexpr EAllowedDOFs operator ~ (EAllowedDOFs inAllowedDOFs)
 }
 
 /// Bitwise OR assignment operator for EAllowedDOFs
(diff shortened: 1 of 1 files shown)


mihe:
I guess there could be some argument for the NOT operator doing `& EAllowedDOFs::All` to ensure that it stays within the expected range, depending on if you look at `EAllowedDOFs` as just a glorified integer or an actual enum.

jrouwe:
Looks good to me. Thanks!

PR #607, opened 2023-07-08
Changed IndexedTriangle::IsDegenerate to include slivers

mihe:
This doesn't have any bearing on godot-jolt/godot-jolt#485, but I was looking through the code and saw an opportunity to maybe simplify things a bit.

I figured the previous `IsDegenerate` check was made more or less redundant by this new cross-product check, so I replaced one with the other. This has the added benefit of also checking for slivers during the sanity checks in `MeshShape::MeshShape` as well as in `Indexify`.


jrouwe:
Hello, I agree that this is a nicer solution. Thanks!

PR #610, opened 2023-07-13
Update README.md - use all cores

Enhex:
on Linux `$(nproc)` evaluates into the number of logical cores (including hyperthreading)


jrouwe:
Thanks!

PR #612, opened 2023-07-13
remove trailing spaces

Enhex:


jrouwe:
Hello, I'll allow it for this file as it serves as a start for other people, but in general the code base is full of trailing white space from before I had an .editorconfig file. I don't want to fix it all because it generates a needless diff on all files.

PR #622, opened 2023-07-19
Remove trailing space characters and ensure EOL for POSIX compliance

aaronfranke:
This PR makes some formatting fixes to this repo, particularly these 2 things:

* Remove trailing space characters. Trailing space characters are useless and do nothing.
* Ensure text files end with a newline character `\n`. This is required for POSIX compliance, which says that text files must be made of lines that end in a newline character. Without this, some tools will not read the files correctly, for example `cat` will not work right. C++ as of C++11 allows non-compliant files but it's still recommended to have files be POSIX compliant, and older compilers [will show warnings](https://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning). GitHub also shows warnings for non-compliant files (the no entry symbol and "No newline at end of file" hover text).

I have VS Code automatically configured to do these things for every file I edit, so not having these issues fixed in the repo is a barrier to me doing other contributions, and creates noisy diffs if I try to edit a file in a fork.

If you want to re-create these changes, copy [this file](https://github.com/godotengine/godot/blob/master/misc/scripts/file_format.sh) and run it on a Linux system, except I also performed these changes on `.bat` files to ensure they don't have trailing spaces and have newlines. If you want, I can set up CI to run this script automatically, it would be easy to just copy-paste the script here and have it run automatically.

I tested running the unit tests and they pass.


mihe:
> I have VS Code automatically configured to do these things for every file I edit, so not having these issues fixed in the repo is a barrier to me doing other contributions, and creates noisy diffs if I try to edit a file in a fork.

I'm probably just stating the obvious here, but with VS Code you can also have workspace-specific settings where you can set `"files.trimTrailingWhitespace": false`, which has worked well enough for me.

This does however mean that `.vscode/settings.json` ends up being listed as an untracked file in Git, but perhaps `.vscode` could be added to `.gitignore` to make that less annoying, seeing as how `.vs` is there already.

jrouwe:
I'm fine with adding .vscode to .gitignore

aaronfranke:
> If you make a meaningful change to another file, then I will accept whitespace fixes / added newlines to that file.

It would be good if you ensure your editor is also set up to fix this, so the problem will indeed disappear over time. If we wait a few years perhaps the diff will be under 1000 lines instead of over 4000 lines. I do think this problem is worth solving eventually, and I can understand your reasons for rejecting even though I disagree.

As for the non-C++ files, indeed there are some files I am not familiar with, like `.tof`, so I'm not sure if they should end in newlines, but there are some I am 100% confident should have newlines, such as shell scripts.

jrouwe:
Ok, I cleaned up the ```.sh``` and ```.bat``` files and added them to the ```.editorconfig``` file. That's not too big of a change.

The ```.tof``` files are written by Jolt's serialization system.

jrouwe:
I've been cleaning whitespace for a while now with every commit and it was causing problems with merging so I bit the bullet and did all of the source files in one go.

aaronfranke:
Superseded by https://github.com/jrouwe/JoltPhysics/pull/688

PR #627, opened 2023-07-21
Fixes a crash caused when fetching a body using an invalid BodyID

AndreaCatania:
Fixes a crash that occurs when fetching a body using an ID that is bigger than the BodyID array size.


jrouwe:
Thanks! I spread out the function over multiple lines now as it was becoming a bit long this way.

PR #635, opened 2023-07-31
-Wdeprecated-copy is clang 13+ and was triggering warnings on earlier clang versions

jrouwe:


joshcoyne:
This works compiling on the Quest 2. Thank you!

PR #640, opened 2023-08-01
Fixes Linux compilation, by removing range loop variable construction.

AndreaCatania:
Fixes the following compilation error:

```
error: loop variable 'c' of type 'const JPH::Ref<JPH::Constraint>&' binds to a temporary constructed from type 'JPH::Constraint*' [-Werror=range-loop-construct]
```


jrouwe:
Thanks!

PR #641, opened 2023-08-02
Add clang-format that disables the automatic formatting.

AndreaCatania:
Add clang-format that disables the automatic formatting.


jrouwe:
Thanks!

PR #643, opened 2023-08-02
Adds selective state save/restore unit tests and more.

AndreaCatania:
This PR introduces the ability to selectively ignore the 'previous delta' and the gravity' when generating the state save and restore AND adds selective state save/restore unit tests.


AndreaCatania:
Please let me know if you prefer to break this test into three different one to make it simpler to follow.

AndreaCatania:
Ok nice.

PR #649, opened 2023-08-04
Failed attempt at fixing the validation function.

AndreaCatania:
The `RestoreState` validation feature is not working properly, I fix what I thought was causing the issue, but after adding some unit tests I found that there is something else. @jrouwe Do you have any clue by reading the below?

---- 

The `SaveState` function is storing the user-provided `EStateRecorderState` into the stream:
```cpp
void PhysicsSystem::SaveState(StateRecorder &inStream, EStateRecorderState inState, const StateRecorderFilter *inFilter) const
{
	inStream.Write(inState);
```

While the `RestoreState` is reading from the stream, and it's providing an hardcoded `EStateRecorderState::None` as initial recorder state:
```cpp
bool PhysicsSystem::RestoreState(StateRecorder &inStream)
{
	EStateRecorderState state = EStateRecorderState::None;
	inStream.Read(state);
```

The issue is that by doing so, the validation is going to always fail as the hardcoded `EStateRecorderState::None` is always going to be different to what's into the stream.

As metioned above, I've tried to fix it by removing it from the stream, but the UnitTests is highlighting that there is still some reading misalignment somewhere.

```
/UnitTests/Physics/PhysicsTests.cpp:1482:
TEST SUITE: PhysicsTests
TEST CASE:  TestSelectiveStateSaveAndRestore

/UnitTests/UnitTestFramework.cpp:49: MESSAGE: Mismatch reading 8 bytes
/UnitTests/UnitTestFramework.cpp:49: MESSAGE: Offset 0: 05 -> 02

Trace/breakpoint trap (core dumped)
```

### DIFF
--- a/Jolt/Physics/PhysicsSystem.cpp
+++ b/Jolt/Physics/PhysicsSystem.cpp
@@ -2373,8 +2373,6 @@ void PhysicsSystem::SaveState(StateRecorder &inStream, EStateRecorderState inSta
 {
 	JPH_PROFILE_FUNCTION();
 
-	inStream.Write(inState);
-
 	if (uint8(inState) & uint8(EStateRecorderState::Global))
 	{
 		inStream.Write(mPreviousStepDeltaTime);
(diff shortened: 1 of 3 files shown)


jrouwe:
Validation cannot work together with partial state saving, the reason why it triggers an assert is that the body manager places the number of bodies in the manager in the variable before reading it, but with a partial save the number you read is always going to be lower so will trigger an assert. There are similar constructs in the other sub systems. I've changed the default from None to All which fixes the problem for a full state save.

PR #662, opened 2023-08-09
Exposes BroadPhase and BodyManager

AndreaCatania:
This commit exposes the internal BodyManager and BroadPhase classes, so it's possible to customize the Physics Engine behaviour.

---- 

This change was needed to implement a custom `SaveState` and `RestoreState`, so that:
- I can avoid using the `BodyManager` to restore the state, as it crashes if the body doesn't exist.
- Allows me to skip networking a Body which state didn't change from the last update.
- Allows me not to network the unnecessary BodyID.

The code I'm using to "Save" an "Restore" is the following one:
```cpp
void JBody3D::jolt_set_state(const Vector<uint8_t> &p_data) {
	if (!body) {
		// The body doesn't exists, so we can just skip it. Eventually the
		// server will make sure to create it.
		return;
	}

	jolt_state.set_data(p_data);
	jolt_state.BeginRead();

	bool is_active;
	jolt_state.ReadBytes(&is_active, 1);

	body->RestoreState(jolt_state);

	if (is_active != body->IsActive()) {
		if (is_active) {
			Jolt::singleton()->body_manager_activate_body(*body);
		} else {
			Jolt::singleton()->body_manager_deactivate_body(*body);
		}
	}

	// Update the Godot transform.
	const Transform3D t(Quaternion(convert(body->GetRotation())), convert_r(body->GetPosition()));
	set_global_transform(t);
}
```

```cpp
Vector<uint8_t> JBody3D::jolt_get_state() {
	if (!body) {
		return Vector<uint8_t>();
	}

	jolt_state.Clear();
	const bool is_active = body->IsActive();
	jolt_state.WriteBytes(&is_active, 1);
	body->SaveState(jolt_state);
	return Vector<uint8_t>(jolt_state.get_data());
}
```

```cpp
void Jolt::body_manager_activate_body(JPH::Body &p_body) {
	JPH::PhysicsSystem *physics_system = worlds[0];
	JPH::BodyManager &bm = physics_system->UnsafeGetBodyManager();
	bm.UnsafeAddBodyToActiveBodies(p_body);
}

void Jolt::body_manager_deactivate_body(JPH::Body &p_body) {
	JPH::PhysicsSystem *physics_system = worlds[0];
	JPH::BodyManager &bm = physics_system->UnsafeGetBodyManager();
	bm.UnsafeRemoveBodyFromActiveBodies(p_body);
}
```

This function is being called when all the bodies states got restored. 
```cpp
void Jolt::world_update_broadphase_aabb(uint32_t p_world_id) {
	JPH::PhysicsSystem *physics_system = worlds[p_world_id];
	JPH::BodyManager &bm = physics_system->UnsafeGetBodyManager();

	// Update bounding boxes for all bodies in the broadphase
	JPH::Array<JPH::BodyID> bodies;
	for (const JPH::Body *b : bm.GetBodies()) {
		if (JPH::BodyManager::sIsValidBodyPointer(b) && b->IsInBroadPhase()) {
			bodies.push_back(b->GetID());
		}
	}

	if (!bodies.empty()) {
		JPH::BroadPhase *bp = physics_system->UnsafeGetBroadPhase();
		bp->NotifyBodiesAABBChanged(&bodies[0], (int)bodies.size());
	}
}
```

### DIFF
--- a/Jolt/Physics/Body/BodyManager.cpp
+++ b/Jolt/Physics/Body/BodyManager.cpp
@@ -863,6 +863,16 @@ bool BodyManager::RestoreState(StateRecorder &inStream)
 	return true;
 }
 
+void BodyManager::UnsafeAddBodyToActiveBodies(Body &ioBody)
+{
+	AddBodyToActiveBodies(ioBody);
+}
+
+void BodyManager::UnsafeRemoveBodyFromActiveBodies(Body &ioBody)
+{
+	RemoveBodyFromActiveBodies(ioBody);
+}
+
 #ifdef JPH_DEBUG_RENDERER
 void BodyManager::Draw(const DrawSettings &inDrawSettings, const PhysicsSettings &inPhysicsSettings, DebugRenderer *inRenderer, const BodyDrawFilter *inBodyFilter)
 {
--- a/Jolt/Physics/Body/BodyManager.h
+++ b/Jolt/Physics/Body/BodyManager.h
@@ -197,6 +197,12 @@ public:
 	/// Restoring state for replay. Returns false if failed.
 	bool							RestoreState(StateRecorder &inStream);
 
+	/// Add a single body to mActiveBodies, note doesn't lock the active body mutex!
+	void							UnsafeAddBodyToActiveBodies(Body &ioBody);
+
+	/// Remove a single body from mActiveBodies, note doesn't lock the active body mutex!
+	void							UnsafeRemoveBodyFromActiveBodies(Body &ioBody);
+	
 	enum class EShapeColor
 	{
 		InstanceColor,				///< Random color per instance
--- a/Jolt/Physics/PhysicsSystem.h
+++ b/Jolt/Physics/PhysicsSystem.h
@@ -135,6 +135,18 @@ public:
 	void						SetGravity(Vec3Arg inGravity)								{ mGravity = inGravity; }
 	Vec3		 				GetGravity() const											{ return mGravity; }
 
+	/// Returns the raw BodyManager without locking it, use `GetBodyLockInterface` instead.
+	inline const BodyManager&	UnsafeGetBodyManager() const								{ return mBodyManager; }
+
+	/// Returns the raw BodyManager without locking it, use `GetBodyLockInterface` instead.
+	inline BodyManager&			UnsafeGetBodyManager()										{ return mBodyManager; }
+
+	/// Returns the raw BroadPhase.
+	inline const BroadPhase*	UnsafeGetBroadPhase() const									{ return mBroadPhase; }
+
+	/// Returns the raw BroadPhase.
+	inline BroadPhase*			UnsafeGetBroadPhase()										{ return mBroadPhase; }
+
 	/// Returns a locking interface that won't actually lock the body. Use with great care!
 	inline const BodyLockInterfaceNoLock &	GetBodyLockInterfaceNoLock() const				{ return mBodyLockInterfaceNoLock; }
 


jrouwe:
Sorry, I don't want to make these classes public. They're internal classes. If I make them public then I will be restricted in the kind of API changes I can do to them.

AndreaCatania:
I understand that and indeed the APIs are marked as `Unsafe` on purpose, so it's expected that these classes can change at any time in any way. We can call them `Internal`, if you think it would be more explicit.

However, restrict access to those classes reduces noticeably the possibility of customizing the physics engine specifically to a game. For example, it's going to be really difficult to customize the `SaveState` and `RestoreState` without maintaining a separate fork, that is going to make the usage experience much worse. Let me highlight that any game will want to optimize and customize the Save/Restore state, as each game will have different networking implementation and requirements.

I'm open to an alternative, though I would prefer to have access to private, classes like these, anyway.

jrouwe:
For me exposing the BodyManager or BroadPhase under a different name doesn't work either. They should just not be accessible.

We already changed the interface for PhysicsSystem::SaveState, what can't you do with it?

> I can avoid using the BodyManager to restore the state, as it crashes if the body doesn't exist.

Why would you send data for a body that doesn't exist on the client. That's just wasting bandwidth right?

> Allows me to skip networking a Body which state didn't change from the last update.

I don't see why this can't be done with the current interface. You get a callback for every body, so you can check for changes there (hashing position/rotation/linear and angular velocity should be good enough as a check).

> Allows me not to network the unnecessary BodyID.

You need some way of identifying which data belongs to which body, how else are you going to do this?

AndreaCatania:
> We already changed the interface for PhysicsSystem::SaveState, what can't you do with it?

Yup, and what we got is really great as I can decide which contact to share.
However, since the bodies sync for a specific client is going to be turned **on** and **off** based on many factors, to be able to sync the physics for a large world, it's not 100% guaranteed that the client has all the info about the bodies that the server is assuming the client knows about.
(From time to time it's expected that a body is not 100% ready)

> Why would you send data for a body that doesn't exist on the client. That's just wasting bandwidth right?

All the bodies that doesn't exist on the client are not being sync.
What I mean is that, when a body goes to sleep (or for any reason its state doesn't change), is going to produce the same state it produced on the previous update: so, the synchronizer is going to detects that and it's going to exclude that state from the delta update.
I call this mechanism delta update: only the changed variables are being sent to the client.

> You need some way of identifying which data belongs to which body, how else are you going to do this?

Yes, exactly, I have already a way to track the body via an ID, that is needed to sync other variables and for this reason I don't need a second ID, it would be just redundant.

----

This is the way I specify which variable to sync, for each body:
```python
	SceneSync.register_variable(character, "movement_scale")
	SceneSync.register_variable(character, "prev_input_velocity")
	SceneSync.register_variable(character, "jolt_sync_data")
```

> ⭐ NOTE: `jolt_sync_data` contains what `jolt_get_state` returns (you can see the code on my first comment).

it would be much much better if I could list all the body variables one by one here, something like this:
```python
	SceneSync.register_variable(character, "movement_scale")
	SceneSync.register_variable(character, "prev_input_velocity")
	SceneSync.register_variable(character, "transform")
	SceneSync.register_variable(character, "body_contacts")
	SceneSync.register_variable(character, "linear_velocity")
	SceneSync.register_variable(character, "angular_velocity")
	SceneSync.register_variable(character, "mass")
	SceneSync.register_variable(character, "can_sleep")
	....
```
The `SceneSync` would be able to detects the changed stuff, and network only the changed variables.
However, this is not extremely important to have TBH, because when a body moves, it's really likely that all the variables serialized by `SaveState` changed.
Though, what is important, is to be able to use this at least `SceneSync.register_variable(character, "jolt_sync_data")`, so I've a much better control on what body is being networked at any given time.

jrouwe:
Hello,

It will not be easy to skip a body from the stream if it hasn't been created on the other side because we don't know if the body is a rigid body or a soft body. The length of the state will differ so we don't know where to continue reading.

What I can imagine is that we add virtual calls like:

* ```virtual void StateRecorder::WriteBodyStart(const BodyID &inID) { Write(inID); }```
* ```virtual void StateRecorder::WriteBodyEnd() { }``` 
* ```virtual BodyID StateRecorder::ReadBodyStart(const BodyID &inExpectedBodyIDForValidation) { BodyID id = inExpectedBodyIDForValidation; Read(id); return id; }``` 
* ```virtual void StateRecorder::ReadBodyEnd() { }``` 

so that you know where each body starts/ends and you can override the virtual to do something else instead of writing the ID to the stream. You can e.g. internally diff the stream of a single body with the previous stream and only send the delta and reconstruct the full stream on the other side. You can add additional information in the stream where bodies start and end so you can skip over the data on the client if the body doesn't exist (you'd just need to check the BodyID in ReadBodyStart and skip over it if it doesn't exist yet). 

This would mean that instead of sending the number of bodies at the beginning of the stream as happens here:

https://github.com/jrouwe/JoltPhysics/blob/0436c32dfd83d26f7559e260378d7fa1d7962841/Jolt/Physics/Body/BodyManager.cpp#L755

```StateRecorder::ReadBodyStart()``` would need to return the invalid BodyID to indicate that the end of the stream has been reached.

I have to say that even without the interface changes you're proposing you can still do with the current interface what you want to do. To get to the broadphase just call:

```static_cast<BroadPhase &>(PhysicsSystem::GetBroadPhaseQuery())```

And if you call ```BodyInterface::ActivateBody``` / ```BodyInterface::DeactivateBody``` before calling ```Body::RestoreState```, then the RestoreState will make sure that the body is in the correct state (if you do it the other way around then ActivateBody will reset the sleep state and you'll get a desync).

**Comment by AndreaCatania** [comment]
Hey Jorrit,

TBH, using a global buffer to sync all the bodies and contacts is not comfortable and nice to use. It doesn't work well with the way the synchronizer works. The Synchronizer must be in control of what it's synchronizing, and should be able to combine the physics properties together with the gameplay variables. Also, since different clients will want a different set of bodies: I want to avoid to generate a different global buffer for each client.

I been thinking about this from yesterday, to find an alternative solution, and after trying and evaluating many approaches I'm pretty convinced that everything we do is going to be complex and will never work for all the games.

I'm convinced that the only way to do it right is to allow people to collect the needed data, as they need: exposing those classes:
```c++
physics_system.__internal.GetBodyManager()
```
So they can really customize the way the stuff is serialized and networked.

The proposed `static_cast<BroadPhase &>(PhysicsSystem::GetBroadPhaseQuery())` is really counterintuitive so a bad way of getting those classes compared to a function with a clear name that is signaling that you are retrieving an internal class.

Anyway, despite I still think the above is the right and easiest way to go, I've two different proposal:

### Proposal 1
Would you like to have these APIs?
```cpp
PhysicsSystem::SaveBodyState(Body* body, StateRecorder &inStream)
PhysicsSystem::RestoreBodyState(Body* body, StateRecorder &inStream)
PhysicsSystem::UpdateBroadphaseAABB()
```

### Proposal 2
Can you change this
https://github.com/jrouwe/JoltPhysics/blob/master/Jolt/Physics/PhysicsSystem.h#L190

to be `protected`? So, I'll be able to override the `PhysicsSystem` to expose those classes on my own?

jrouwe:
Ok, I added a unit test and I'll merge it when it goes through the build. Closing this PR.

AndreaCatania:
Lovely, thx!

PR #668, opened 2023-08-11
Allowed for moving triangle data into MeshShapeSettings

mihe:
It's currently not possible to hand over ownership (through `std::move`) of `VertexList`, `IndexedTriangleList` or `PhysicsMaterialList` to `MeshShapeSettings`, due to it taking these parameters as `const&`, leading to it always copying said data.

This PR fixes that by instead taking the triangle data by-value and moving it into each respective member.

For the case where you're not handing off ownership of the data, this will potentially incur an extra (presumably negligable) move constructor invocation for each member, but for the case where you do hand off the data, you will now avoid having to copy the data.

### DIFF
--- a/Jolt/Physics/Collision/Shape/MeshShape.cpp
+++ b/Jolt/Physics/Collision/Shape/MeshShape.cpp
@@ -70,7 +70,15 @@ static JPH_INLINE const TriangleCodec::TriangleHeader *sGetTriangleHeader(const 
 	return inTree.Get<TriangleCodec::TriangleHeader>(NodeCodec::HeaderSize);
 }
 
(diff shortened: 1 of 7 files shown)


jrouwe:
Hello,

I've modified the change a bit because if you forgot to use std::move or had a const reference to a list you would actually make an additional copy compared to the previous situation.

I also made use of the new constructors where possible in the samples.

jrouwe:
Hold on, I think I'm actually talking BS, let me test this again.

jrouwe:
Ok, I see now how you intended this to work. I need to see if this is compatible with the JavaScript version of Jolt.

PR #675, opened 2023-08-18
mark parameters with [[maybe_unused]] in StateRecorderFilter methods

NicSavichev:
to avoid warnings (treated as errors for some compilation settings) Jolt/Physics/StateRecorder.h:35:60: error: unused parameter ‘inBody’ [-Werror=unused-parameter]


jrouwe:
Thanks, and sorry for forgetting this.

PR #701, opened 2023-09-11
Adds functions to check which SIMD operations are supported.

AndreaCatania:
These functions are useful to verify that the generated jolt ABI is compatible with the one that the parent project expects: this is important especially in the case the Jolt lib and the parent projects are compiled with two different build tools.


jrouwe:
Hello,

This check is already implemented here:

https://github.com/jrouwe/JoltPhysics/blob/60e16e44e8214eff5a9b9c0485d44d637ce46ee6/Jolt/RegisterTypes.cpp#L75-L79

And here:

https://github.com/jrouwe/JoltPhysics/blob/60e16e44e8214eff5a9b9c0485d44d637ce46ee6/Jolt/Core/Core.h#L12-L61

For the ABI it actually doesn't matter which instruction sets you compiled with.

AndreaCatania:
Thanks for the links, I've improved this PR based on your last message.

First let me clarify why I needed to add the function `VerifyJoltVersionId()`. Since in godot each module is compiled with its own set of macros, and I've to specify the same macros across two modules manually: I needed a way to verify that the module **B**, which is using the jolt library compiled by the module **A**, is using the correct ABI.

In the above case, the `RegisterTypes()` function is executed by the module **A**, so I needed an extra warranty for the module **B**.

### Notice 1
I tried to use the same approach/style used by `RegisterTypes`, but on the module **B** the function was never failing.
```cpp
JPH_EXPORT extern bool VerifyJoltVersionIdInternal(uint64);
inline bool VerifyJoltVersionId() { return VerifyJoltVersionIdInternal(JPH_VERSION_ID); }
```
I think the issue was that the function wasn't inlined, and the module **B** was using the function compiled somewhere by the module **A**. The solution to this problem was to let the module call he function explicitly:
```cpp
CRASH_COND(!JPH::VerifyJoltVersionId(JPH_VERSION_ID));
```

### Notice 2
I've changed `uint64` to `std::uint64_t` so I can easily use `JPH_VERSION_ID` from outside the jolt library (the compiler was complaining that JPH::uint64 was unknown).

### Notice 3
I've added some more stuff into the `JPH_VERSION_ID` that I think are important to check. Not specifying those macros cause the application to have crashes due to undefined behaviour.

- `JPH_ENABLE_ASSERTS`
- `JPH_FLOATING_POINT_EXCEPTIONS_ENABLED`

These are not mandatory, as `Core.h` are defining it.
However, I feel like we should check these macros anyway, because the jolt library and the parent applications are compiled with two different build tools that may have a different config and so may have a different set of SIMD operations enabled.
Have a way to detect that, doesn't sounds like a bad idea, especially considering that these are defining the base types of some classes: `Jolt/Math/MathTypes.h::21` `Jolt/Math/DVec3.h::19-31`  
- `JPH_USE_AVX`
- `JPH_USE_SSE`
- `JPH_USE_NEON`
If you think that's not necessary as they are not causing UB even if they are different, I'll remove it.

jrouwe:
> inline bool VerifyJoltVersionId() { return VerifyJoltVersionIdInternal(JPH_VERSION_ID); }

Can you try this with JPH_INLINE instead? (this will use 'force inline').

> I've changed uint64 to std::uint64_t so I can easily use JPH_VERSION_ID from outside the jolt library (the compiler was complaining that JPH::uint64 was unknown).

JPH_VERSION_ID is defined in the same file as JPH::uint64 so I don't really see why this should cause a compiler issue. Can you take another look at this? For consistency I'd like everything to use uint64 instead of the std types.

About the name ```VerifyJoltVersionId```: Can this be ```VerifyJoltVersionID``` to be consistent with naming used elsewhere in the library? And can you move the function to RegisterTypes.h/cpp (otherwise it will be defined in every compilation unit and that's not really needed)?

> JPH_ENABLE_ASSERTS

This is indeed one that I missed that should be added, thanks for that!

> JPH_FLOATING_POINT_EXCEPTIONS_ENABLED

This is now checked 2x (bit 3 and 11)

> If you think that's not necessary as they are not causing UB even if they are different, I'll remove it.

I don't think this check is needed as the classes are binary compatible with/without AVX/SSE/NEON.

Finally, can you move the code block that defines JPH_VERSION_ID back to where it was?

AndreaCatania:
All points are done.

- The `JPH_INLINE` fixed the issue and I was able to implement `VerifyJoltVersionID();`.
- I needed to add `#include <Jolt/Core/Core.h>` into `RegisterTypes.h` where `JPH_INLINE` is defined.

AndreaCatania:
Hello @jrouwe
Do you think this is good to merge?

jrouwe:
Yes, sorry, I was AFK yesterday. 

I made a tiny fix: Core.h is included through Jolt.h which you should always include before including anything else (quirky I know, but it was meant so that you can put Jolt.h in a precompiled header as described [here](https://github.com/jrouwe/JoltPhysics/blob/master/Build/README.md)).

AndreaCatania:
Oh, I see I'll keep that in mind, thanks!

PR #738, opened 2023-11-03
Changed AlignedAllocate to default to posix_memalign

mihe:
I'm in the process of implementing iOS support for Godot Jolt, and it turns out that `aligned_alloc` is only available on iOS >=13, as seen [here](https://opensource.apple.com/source/libmalloc/libmalloc-317.140.5/include/malloc/_malloc.h.auto.html). According to that header it also appears that `aligned_alloc` is not available on macOS versions prior to 10.15 (Catalina).

What is available however (also found in the link above) is `posix_memalign`.

I figured instead of adding even more platform ifdef's to `AlignedAllocate` I would just simplify things a bit and always default to `posix_memalign` on anything but Windows, as this seems to be available on Android as well. Although, I'm not sure how this change fares on that one blue platform.

I ran `PerformanceTest` on macOS and saw no difference compared to `aligned_allocate`.

Also note that `posix_memalign` technically returns an error code, but seeing as how none of the other allocation functions had any error-handling I opted to just ignore it.

### DIFF
--- a/Jolt/Core/Memory.cpp
+++ b/Jolt/Core/Memory.cpp
@@ -36,7 +36,7 @@ JPH_ALLOC_SCOPE void *JPH_ALLOC_FN(AlignedAllocate)(size_t inSize, size_t inAlig
 	return _aligned_malloc(inSize, inAlignment);
 #else
 	void *block = nullptr;
-	(void)posix_memalign(&block, inAlignment, AlignUp(inSize, inAlignment));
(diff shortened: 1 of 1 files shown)


jrouwe:
Note that I just fixed the Shared library compiler errors as part of #739 (github must have bumped a compiler version).

mihe:
Alright, I guess you can merge in master here once #739 has been merged, if you'd prefer to see a successful run first.

jrouwe:
Thanks!

PR #749, opened 2023-11-09
Enforce MustIncludeC in GetClosestPointOnTriangle

bttner:
This PR is referring to the discussion https://github.com/jrouwe/JoltPhysics/discussions/740#discussioncomment-7515585 and fixes the described problem.

Note that we can overwrite the variable _ab_ as the function returns after testing the edge BC and, therefore, its last use only involves testing the edge AB.

### DIFF
--- a/.github/workflows/determinism_check.yml
+++ b/.github/workflows/determinism_check.yml
@@ -2,7 +2,7 @@ name: Determinism Check
 
 env:
     CONVEX_VS_MESH_HASH: '0x10139effe747511'
-    RAGDOLL_HASH: '0xf6e41ad9d35ffa7f'
+    RAGDOLL_HASH: '0xcdcbb4da185d1a13'
 
 on:
   push:
--- a/.github/workflows/determinism_check.yml
+++ b/.github/workflows/determinism_check.yml
@@ -2,7 +2,7 @@ name: Determinism Check
 
 env:
     CONVEX_VS_MESH_HASH: '0x10139effe747511'
-    RAGDOLL_HASH: '0xfb18854efbb36f83'
+    RAGDOLL_HASH: '0xf6e41ad9d35ffa7f'
 
 on:
   push:
--- a/Jolt/Geometry/ClosestPoint.h
+++ b/Jolt/Geometry/ClosestPoint.h
@@ -183,6 +183,29 @@ namespace ClosestPoint
 			Vec3 closest_point = inC;
 			float best_dist_sq = inC.LengthSq();
 
+			// If the closest point must include C then A or B cannot be closest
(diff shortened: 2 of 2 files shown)


jrouwe:
I've taken a quick look at the failing unit tests. These don't mean that the collision detection code is wrong but it means that the collision response changed enough that the penetration between a simulated box and the floor is slightly larger than before. I will take a look at it, but it is going to take me some time to verify the new behavior.

bttner:
Unfortunately, the Unit test "TestLinearCastBoxVsInactiveDiscreteBoxAngled" fails.

When these lines
https://github.com/jrouwe/JoltPhysics/blob/b9bdabe8152cefcf4ca4826990d2c7f57a2094cd/UnitTests/Physics/MotionQualityLinearCastTests.cpp#L306-L310
become
```
CHECK_APPROX_EQUAL(box1.GetLinearVelocity(), new_velocity1, 1.0e-3f);
CHECK_APPROX_EQUAL(box1.GetAngularVelocity(), Vec3::sZero(), 2.0e-3f);
CHECK_APPROX_EQUAL(box2.GetPosition(), RVec3::sZero());
CHECK_APPROX_EQUAL(box2.GetLinearVelocity(), new_velocity2, 1.0e-3f);
CHECK_APPROX_EQUAL(box2.GetAngularVelocity(), Vec3::sZero(), 2.0e-3f);
```
then the mentioned test passes. So the accuracy is worse by a factor of 10.

What is odd is that this test calls GetClosestPointOnTriangle with MustIncludeC = false. I am now investigating why that is. The test should not fail.

bttner:
The problem is the following.

In the mentioned test MustIncludeC is false and swap_ac are true. The distance to edge AB is (almost) the same as to Edge AC.
- Before the change, edge AC is tested before edge AB (the block with comment // Edge BC), which is why edge AC is chosen.
- After the change, edge AB is tested before edge AC, which is why edge AB is chosen

The closest point to edge AB is (0.1, 0, 5.96e-08), while the closest point to edge AC is calculated as (0.1, 0, 0).

bttner:
I changed the order of testing the edges, i.e., edge AC and BC are getting prioritized now. This change should pass all unit tests.

Moreover, even if MustIncludeC is false, the closest point can be expected to be most of the time on the edge AC, BC, or inside. So it makes sense to test these cases first.

jrouwe:
I've been testing with this change and there was one issue with it: If we test AC before A and the closest point is A, we will find that ```v == 0``` and calculate ```best_dist_sq = |A|^2```. If we then later test vertex A we find ```!(a_len_sq < best_dist_sq)``` (because they're equal) so while the returned position is correct, the returned outSet is not (it return 0b101 instead of 0b001). For CastShape this results in more vertices being used in GJK and a slight slowdown.

jrouwe:
Thanks for your contribution btw!

PR #751, opened 2023-11-09
Alternative approach for supporting Android

jrouwe:


mihe:
It seems I had misread one of your `set(CMAKE_CXX_FLAGS ...)` calls as actually overwriting it instead of appending/prepending to it, so the flags mentioned in #747 were actually not being touched in the first place, as they were being added through `CMAKE_CXX_FLAGS`.

But it seems like your changes are still of some relevance, as there was a flag being brought in through `CMAKE_CXX_FLAGS_DEBUG` that might be useful, which was `-fno-limit-debug-info`, and who knows what else might be added in the future.

I guess there's also the argument that not having this flag sets you up for problems similar to #319, where you could potentially be stomping user-provided `CMAKE_CXX_FLAGS_<CONFIG>` as well.

Frankly I would perhaps even argue for this flag not being a thing at all and instead do some string-replacing to tweak whatever settings that you feel have to be tweaked, like replacing `-O2` with `-O3` if that's important. But I can understand the need/want to be in full control of things like optimization flags.

In general the compilation flags tend to not change so long as you don't bump the minimum CMake version for your project (with Android being a notable exception I guess) so my personal approach has mostly been to rely on whatever flags are generated by CMake for that particular minimum CMake version and then add/tweak flags only as necessary, preferably through whatever `CMAKE_MSVC_*` (and similar) variables that might be available. Then once (if ever) you bump the minimum CMake version you'd go through them again and make sure it's all good, and maybe make use of any new `CMAKE_*` variables that might have been added.

There's also the question of why this flag doesn't apply to MSVC-like builds I guess.

Anyway, that's all mostly just nitpicking and matters of preference. This change is still an improvement on #747, so I'd be happy to see it be merged.

jrouwe:
Ok, thanks! I'll merge it.

PR #752, opened 2023-11-10
Fix Linux shared compilation

SirLynix:
(Looks like I created this PR on the main repo instead of my fork, sorry about that, I'm making this a draft until I fix this)

Jolt doesn't compile as a shared library on Linux using GCC (works fine with Clang), because of a weird parsing conflict between the `[[nodiscard]]` and the visibility attribute on Color and OrientedBox classes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,6 +74,26 @@ jobs:
       working-directory: ${{github.workspace}}/Build/Linux_${{matrix.build_type}}_${{matrix.clang_version}}
       run: ctest --output-on-failure
 
+  linux-gcc-so:
+    runs-on: ubuntu-latest
+    name: Linux GCC Shared Library
+    strategy:
+        fail-fast: false
+        matrix:
+            build_type: [Debug, Release, Distribution]
+            clang_version: [g++]
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v4
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/Build/Linux_${{matrix.build_type}}_${{matrix.clang_version}} -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_CXX_COMPILER=${{matrix.clang_version}} -DBUILD_SHARED_LIBS=Yes Build
+    - name: Build
+      run: cmake --build ${{github.workspace}}/Build/Linux_${{matrix.build_type}}_${{matrix.clang_version}} -j 2
+    - name: Test
+      working-directory: ${{github.workspace}}/Build/Linux_${{matrix.build_type}}_${{matrix.clang_version}}
+      run: ctest --output-on-failure
+
   msys2_mingw_gcc:
     runs-on: windows-latest
     defaults:
--- a/Jolt/Core/Color.h
+++ b/Jolt/Core/Color.h
@@ -12,7 +12,7 @@ class Color;
 using ColorArg = Color;
 
 /// Class that holds an RGBA color with 8-bits per component
-class [[nodiscard]] JPH_EXPORT Color
+class [[nodiscard]] JPH_EXPORT_TYPE Color
 {
 public:
 	/// Constructors
--- a/Jolt/Core/Color.h
+++ b/Jolt/Core/Color.h
@@ -12,7 +12,7 @@ class Color;
 using ColorArg = Color;
 
 /// Class that holds an RGBA color with 8-bits per component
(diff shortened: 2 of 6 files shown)


SirLynix:
Now it's ready, here's a better description:

There is a parsing issue when C++17 attributes `[[something]]` and GCC 
attributes `__attribute(something)__` appears in a class specification.

This can be fixed by using C++17 attributes syntax for GCC attributes.
However, this isn't a drop-in replacement for every usage.

I think it's good now, I couldn't find a change impacting less files or not requiring a separate macro (`[[gnu::visibility("default")]]` and `__attribute(visibility("default"))__` don't seem to be strictly equivalent.

SirLynix:
Just to be sure you noticed this PR is ready @jrouwe (sorry for the notification)

jrouwe:
Yes, I did notice this but I wanted to take a good look at the change.

It looks like the issue has been fixed in gcc 13 (see: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69585) and when trying this out on [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'class+%5B%5Bnodiscard%5D%5D+__attribute__((visibility(%22default%22)))+Test1+%7B+%7D%3B%0A%0Aclass+alignas(16)+__attribute__((visibility(%22default%22)))+Test2+%7B+%7D%3B'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g131,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.1+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+clang+(trunk)',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+13.1+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) it indeed appears to be fixed.

So if you can use gcc 13+ then that would be great! (no fixes needed)

If you can't use gcc 13 then my feedback is that I'm not too happy with the new ```JPH_EXPORT_TYPE``` macro as it is really confusing when it is needed (there are 100's of types in Jolt but only 4 types use it), perhaps we should rename the macro to ```JPH_EXPORT_GCC_BUG_WORKAROUND``` and referring to the bug tracker URL in the comments above the macro definition. Then at least it is clear that it is a hack for a bug in gcc instead of an official macro that should be used everywhere.

SirLynix:
> So if you can use gcc 13+ then that would be great! (no fixes needed)

Unfortunately Github actions and Debian are on GCC <13.

> If you can't use gcc 13 then my feedback is that I'm not too happy with the new JPH_EXPORT_TYPE macro as it is really confusing when it is needed (there are 100's of types in Jolt but only 4 types use it), perhaps we should rename the macro to JPH_EXPORT_GCC_BUG_WORKAROUND and referring to the bug tracker URL in the comments above the macro definition. Then at least it is clear that it is a hack for a bug in gcc instead of an official macro that should be used everywhere.

I initially wanted to use JPH_EXPORT_TYPE on every class but that was a bigger change (I still can do it if you want).

I'm also okay with JPH_EXPORT_GCC_BUG_WORKAROUND, as it would only appear on three types, and this workaround could be removed once GCC 13 becomes mainstream, I'll make that change.

Thanks for the feedback!

jrouwe:
Thanks for the contribution!

PR #754, opened 2023-11-11
In architecture.md, Fix a typo to be closer to intent on "driving to an angle" ignoring force limits

duchainer:
We repeated twice position, but it would make more sense if we meant that "driving to an angle" doesn't use the force limits, only the torque limits.

Let me know if anything more needs to be done to this PR, but feel free to use it and change it as you wish.


jrouwe:
Thanks, that was indeed wrong!

PR #760, opened 2023-11-15
Tautology in GetPenetrationDepthStepEPA

bttner:
The function ConvexHullBuilder::AddPoint calls ConvexHullBuilder::FindEdge, which sets `inFacingFace->mRemoved = true`. Thus, the condition `if (t->mRemoved)` in GetPenetrationDepthStepEPA is always true.

Furthermore, this PR also fixes a "misleading" comment in TriangleQueue::PopClosest: Given sTriangleSorter, the function pop_heap moves the closest point to the end, not the largest one.


jrouwe:
Thank you!

PR #767, opened 2023-11-21
Support for compiling with ninja on Windows

jrouwe:


mihe:
Just to clarify, the issue here isn't with Ninja per se and more of an issue with compiling with the Windows-native version of `clang++`, as opposed to something like `clang-cl`. I've been using Ninja on all platforms for Godot Jolt since the beginning, albeit with `clang-cl` on Windows.

Anyway, this doesn't seem to work quite right just yet. You'll need to use `if (MSVC)` in these three places as well:

https://github.com/jrouwe/JoltPhysics/blob/d0a5839939f5b48335a905e10de7b0b43574b0b0/Build/CMakeLists.txt#L234-L236

https://github.com/jrouwe/JoltPhysics/blob/d0a5839939f5b48335a905e10de7b0b43574b0b0/Build/CMakeLists.txt#L260-L262

https://github.com/jrouwe/JoltPhysics/blob/d0a5839939f5b48335a905e10de7b0b43574b0b0/Build/CMakeLists.txt#L271-L273

Then there's also the problem of what to do with the `-pthread` flag, since that one doesn't make a whole lot of sense when compiling on Windows, which leads Clang to emit a warning about it, which becomes an error.

I suppose you can choose to either not add that flag when `"${CMAKE_SYSTEM_NAME}" STREQUAL "Windows"`, or just throw in `-Qunused-arguments` when `"${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"`, much like what's already being done for `clang-cl` on Windows:

https://github.com/jrouwe/JoltPhysics/blob/d0a5839939f5b48335a905e10de7b0b43574b0b0/Build/CMakeLists.txt#L126

With those changes I can successfully build and run all of Jolt's applications.

mihe:
It struck me just now that maybe MinGW still needs the `-pthreads` flag? I'm not very familiar with how the runtime stuff work there.

jrouwe:
Apparently not:
image

PR #773, opened 2023-11-26
Fixed AllowedDOFs comment

Scorr:
Heyo, I think this should be correct.


jrouwe:
Yes, thank you for spotting that!

PR #785, opened 2023-12-06
Changed the meaning of the number of Constraint::mNumVelocity/PositionStepsOverride

jrouwe:


mihe:
Just so I understand this change correctly...

> This means that if all constraints in an island have a lower value than the default, we will now use the lower value instead of the default. This allows simulating an island at a lower precision than the default.

> Bodies can now also override the default number of solver iterations. This value is used when the body collides with another body and a contact constraint is created

I assume that contact constraints are considered part of the constraint islands mentioned in the first quote?

If so, if we have a scenario where the default/global iterations are set to something like 16, and body A and body B are constrained by joint C, and joint C has both its overrides set to 1, while the bodies don't have any overrides set, and then body A comes in contact with body B, will the iterations for that entire island go back to being 16?

jrouwe:
> I assume that contact constraints are considered part of the constraint islands mentioned in the first quote?

Correct.

> If so, if we have a scenario where the default/global iterations are set to something like 16, and body A and body B are constrained by joint C, and joint C has both its overrides set to 1, while the bodies don't have any overrides set, and then body A comes in contact with body B, will the iterations for that entire island go back to being 16 for the duration of that contact?

Correct too.

PR #791, opened 2023-12-12
Adds a tool to validate the VirtualCharacter determinism, to find the code that is making it non deterministic.

AndreaCatania:
Hello, I've debugged a desync issue on the game [The Mirror](https://www.themirror.space/), that occurs when the Virtual Character leaves the ground (OR touches the ground for the first time). Which is easily reproducible by jumping.

ezgif com-video-to-gif-converted

The purpose of this code is to find the route cause of this desync, which seems coming from Jolt, and give more tools to the Jolt's community to keep the code deterministic.

On the video below you will see the `VirtualCharacter` replaying a series of input, 4 times, recorded just few seconds before - offscreen. You can notice, from the array printed on the line **NON DETERMINISTIC FRAMES**, that even just moving the capsule horizontally on a flat plane (not even jumping) the taken snapshots differ, meaning the execution was not deterministic.
My educated guess is that there is some non deterministic code inside the `VirtualCharacter`, which result is much noticeable when the `VirtualCharacter` leaves or touches the ground.

https://github.com/jrouwe/JoltPhysics/assets/8342599/4e922935-2b97-4488-8a4f-091884915cd1

If you come up with something regarding this issue that can help me fixing it, please let me know.

# Change description
This commit introduce a basic rewinding mechanism, that allows to record a series of input and replay them. This mechanism can be used to test the VirtualCharacter determinism on many situations.

When playing, is now possible to press `Y` to start recording the inputs (press `Y` again to stop it), and then by pressing `U` it's possible to replay those inputs and validate the VirtualCharacter state on each frame. On the UI, will be displayed the list of frames which the state differs from the one originally recorded so the rewinding was not deterministic.

### DIFF
--- a/Samples/Tests/Character/CharacterBaseTest.cpp
+++ b/Samples/Tests/Character/CharacterBaseTest.cpp
@@ -529,11 +529,8 @@ void CharacterBaseTest::Initialize()
 	}
 }
 
-void CharacterBaseTest::PrePhysicsUpdate(const PreUpdateParams &inParams)
+void CharacterBaseTest::FetchNewInput(const PreUpdateParams &inParams, Vec3Arg& outMovementDirection, bool &outJump, bool &outSwitchStance)
 {
-	// Update scene time
-	mTime += inParams.mDeltaTime;
-
 	// Determine controller input
 	Vec3 control_input = Vec3::sZero();
 	if (inParams.mKeyboard->IsKeyPressed(DIK_LEFT))		control_input.SetZ(-1);
@@ -560,6 +557,21 @@ void CharacterBaseTest::PrePhysicsUpdate(const PreUpdateParams &inParams)
 		else if (key == DIK_RCONTROL)
 			jump = true;
 	}
+
+	outMovementDirection = control_input;
+	outJump = jump;
+	outSwitchStance = switch_stance;
+}
+
+void CharacterBaseTest::PrePhysicsUpdate(const PreUpdateParams &inParams)
+{
+	// Update scene time
+	mTime += inParams.mDeltaTime;
+
+	Vec3 control_input = Vec3::sZero();
+	bool jump = false;
+	bool switch_stance = false;
+	FetchNewInput(inParams, control_input, jump, switch_stance);
 
 	// Animate bodies
 	if (!mRotatingBody.IsInvalid())
@@ -684,5 +696,5 @@ void CharacterBaseTest::DrawCharacterState(const CharacterBase *inCharacter, RMa
 	const PhysicsMaterial *ground_material = inCharacter->GetGroundMaterial();
 	Vec3 horizontal_velocity = inCharacterVelocity;
 	horizontal_velocity.SetY(0);
-	mDebugRenderer->DrawText3D(inCharacterTransform.GetTranslation(), StringFormat("State: %s\nMat: %s\nHorizontal Vel: %.1f m/s\nVertical Vel: %.1f m/s", CharacterBase::sToString(ground_state), ground_material->GetDebugName(), (double)horizontal_velocity.Length(), (double)inCharacterVelocity.GetY()), Color::sWhite, 0.25f);
+	mDebugRenderer->DrawText3D(inCharacterTransform.GetTranslation(), StringFormat("State: %s\nMat: %s\nHorizontal Vel: %.1f m/s\nVertical Vel: %.1f m/s\n%s", CharacterBase::sToString(ground_state), ground_material->GetDebugName(), (double)horizontal_velocity.Length(), (double)inCharacterVelocity.GetY(), GetAdditionalCharacterStateInfo()), Color::sWhite, 0.25f);
 }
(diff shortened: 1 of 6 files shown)


jrouwe:
Hello,

I tried to reproduce your problem, but I think the reason why you're seeing non-deterministic frames is because of this line:

https://github.com/jrouwe/JoltPhysics/blob/cffc45188439481a8269fedd8747faedf1444c8c/Samples/Tests/Character/CharacterVirtualTest.cpp#L99

You're not storing mDesiredVelocity in your FrameSnapshot object which means the input is not deterministic. If I set ```sEnableCharacterInertia = false``` then I no longer get non-deterministic frames.

About your idea of making it easier to detect non-deterministic behavior: I think the idea of storing keyboard input is great, but the implementation is only storing the character, which means that any interaction with a non-static physics object will cause non-deterministic frames. I think a better implementation would be to store any relevant key values that were read in CharacterVirtualTest::SaveState and restore them in CharacterVirtualTest::RestoreState. If we detect that we're playing back through the existing playback functionality in SampleApp:

https://github.com/jrouwe/JoltPhysics/blob/cffc45188439481a8269fedd8747faedf1444c8c/Samples/SamplesApp.h#L130-L144

we simply don't overwrite the key values with the values from mKeyboard but use the restored ones. This way you get recording / determinism validation of both the character and all physics objects. I can give it a shot to implement this if you want, should be pretty simple.

jrouwe:
Turns out it was not quite as simple as I imagined it to be, got something half working. Will continue tomorrow.

AndreaCatania:
I've added `mDesiredVelocity` and that fixes the sync issue as you suggested; so the desync issue I'm debugging doesn't come directly from jolt, which is good to know.

Regarding the idea of restoring all the objects state, I didn't want to restore the state for all the objects on purpose, to make sure that we can partially restore the scene state without too many issues.
This is crucial, because it's unlikely that a game would restore the state for all the simulating objects. Usually, the game's scene contains many objects and not all the simulating objects are important for the networked simulation; be forced to restore the state for the entire scene would result in a too big snapshot; too big to efficiently network it.

For the above reason, it's important that Jolt allows to restore the physics state partially, for the needed objects, and having a test / sample to validate just the `VirtualCharacter` is great.

#### Proposal 1

The issue you've mention (about the character touching non static objects would result in a non deterministic simulation) is real, and I think that an alternative way of solving it is: to show a RED warning when the character touches a non static object while recording. In this way, the user is aware about it.

#### Proposal 2

In alternative, a setting to selectively choose whether the entire scene or just the `VirtualCharacter` is restored, would be enough to test the above.

jrouwe:
While I think it should indeed be possible to partially save the state, I don't think this makes sense for the character demos as that means they will report non-determinism as soon as you touch something non-static (because everything is relevant in those demos as you can touch all of the objects) and that will get me bug reports.

I've created a way to save the input state in the character test (and I will extend it to the vehicle tests too):

https://github.com/jrouwe/JoltPhysics/tree/feature/save_player_input

This led me to find a small problem in the state that was saved in that demo, but it also highlights a determinism problem: When you turn on 'check determinism' in the sample app menu, walk up to the 3 stacked boxes and start pushing them the validation fails. For some reason the order of mActiveContacts is not deterministic which will lead to a non-deterministic simulation. I'll continue later with the investigation.

AndreaCatania:
Yeah, I agree, better to make a proper tool to test partial state recording, instead to rely on one that can report false positives.
I'm quite happy that you already found a bug, and I'm looking forward to try what you are implementing :) 

Feel free to close this PR as soon as you merge yours, and once again thanks for making Jolt!

jrouwe:
I found something and fixed it in https://github.com/jrouwe/JoltPhysics/pull/795

I'm not sure however that this is affecting you as the constraints were sorted afterwards anyway. Only calls to ```CharacterVirtual::GetActiveContacts()``` would receive the contacts in a different order. Can you test if this change makes a difference?

There is a problem with this fix, if you exceed the max number of hits, the hits that it throws away do not have to be deterministic and it is really hard to fix this and stay performant at the same time. Can you check that you're not hitting this code path by checking ```CharacterVirtual::GetMaxHitsExceeded() == false```?

AndreaCatania:
I've found the issue on my side and fixed it. Jolt wasn't the issue, but something else on the app side. However, as soon as I'm on my computer I can test your change and tell you if it makes any difference.

jrouwe:
Ok, in the mean time I finished player input recording in #796. I'm closing this PR, if you find anything then add a comment here or create a new ticket.

AndreaCatania:
Hey @jrouwe I've reverted the desync fix I did on the app, to test whether this https://github.com/jrouwe/JoltPhysics/pull/795 made any difference.
Your change fixed the issue. My initial suspect about the contacts being reported in a non deterministic manner, causing a desync when the capsule start/stop touching the ground, was correct; I was discouraged after seeing the sample working fine 😅 TBH.

Worth a mention that on the app side, prior to your change https://github.com/jrouwe/JoltPhysics/pull/795, I've fixed the issue by re-implementing the character processing by strictly following the implementation you did on the sample project. That implementation doesn't trigger the desync.

In any case, that's great your were able to find and **fully** fix the issue. Thank you!

PR #820, opened 2023-12-26
Update README.md

cuppajoeman:
Added section about documentation


jrouwe:
Hello,

Thanks for your PR, but I feel this is a bit of redundant information as the [Build](https://github.com/jrouwe/JoltPhysics/blob/master/Build/README.md) section already mentions how to build with Doxygen and the link to the API documentation is a couple of paragraphs above.

cuppajoeman:
> Hello,
> 
> Thanks for your PR, but I feel this is a bit of redundant information as the [Build](https://github.com/jrouwe/JoltPhysics/blob/master/Build/README.md) section already mentions how to build with Doxygen and the link to the API documentation is a couple of paragraphs above.

The instructions in the build only show how you can build on windows for linux users following those directions wouldn't work.  It's also at the very bottom and I only found that section because you explicitly mentioned it here. 

Also I contest the fact that it's in Build section, because a contributor who wants to add documentation probably wouldn't expect that information to be found in the build section of a repository because in their mind they want to contribute. 

Knowing that this info would be in the build section implicitly assumes that you know you actually *Build* documentation, which isn't true in general because we can just edit the main readme file do a pr and it's updated. You would only know we also *build* documentation if you've already used doxygen and know how it operates.

It's also unclear initially what files are being used to build the doxygen documentation because they use markdown extensions which makes them appear as regular markdown files, not inputs to doxygen that will eventually be built. The content of the paragraph I wrote mentioned this and clears up where the inputs come from.

In general I feel like a paragraph explaining our documentation set up and explaining that we're using github pages to automatically construct the docs and host them every time a change occurs there would be welcome to any beginner who wants to help improve the docs.

Let me know what you think about my comments here. Maybe we can decide on a way to combine what we have in the Build section and what I wrote in this pull request somehow?

PR #821, opened 2023-12-26
Update README.md

cuppajoeman:
format build section in such a way so that the user is not bombarded by irrelevant build instructions to them. 

Instead they can access the relevant build instructions through at most 2 clicks.


jrouwe:
Thank you!

PR #823, opened 2023-12-27
Add Linux support for SetThreadName

thendrix:
Quick patch done over xmas vacation to debug high thread count in an app.


jrouwe:
Thank you!

PR #835, opened 2023-12-31
Use SetThreadDescription on Windows if supported

SirLynix:
This allows MinGW to have thread names starting from Windows 10 version 1607 with [SetThreadDescription](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription), and allows the debugger to retrieve thread name even if it was attached after starting the program (or in a dump).


jrouwe:
Thanks!

PR #843, opened 2024-01-01
Improve Doxygen Documentation

cuppajoeman:
Added a proper index page to the doxygen docs and a build page.

https://github.com/jrouwe/JoltPhysics/assets/19198104/2dbbf919-f225-4f62-9c29-b15a647fe6ee

I also enabled TOC's on certain pages and made it so that sections are automatically shown in the treeview for easier navigation.

I think that the workflow for learning about Jolt is a little nicer when using doxygen as given any page you can first use the table of contents as a quick way to see the overview of the page, and then the treeview on the left to navigate after that point.

I think it's also useful to be able to link directly to things via doxygen such as source files and line numbers. Also when working offline we still have access to documentation and add to it.

I guess this brings up the point of comparison between github documentation and the doxygen documentation, and how further development should happen with respect to the docs.

From what I can see it seems like upon initial visitation of the Jolt repository users will first see the docs on github, and then for further details they can go to the doxygen documentation, so maybe this pull request isn't the right approach. 

Do you think it's best to encourage users towards one place for all documentation or keep some split on github and some on doxygen?


jrouwe:
Hello,

The new layout is nice, but copy pasting the index page and build information makes it very difficult for me to maintain the documentation as I need to update things in 2 places. Can you please remove the duplication?

cuppajoeman:
> The new layout is nice, but copy pasting the index page and build information makes it very difficult for me to maintain the documentation as I need to update things in 2 places. Can you please remove the duplication?

Hey @jrouwe I agree having to maintain the documentation in two places would be bad.

The reason I duplicated things here is that doxygen supports a superset of the regular markdown that github supports, so if I augmented our regular documentation it would leave artifacts that doxygen is picking up on.

The first example would the command `[TOC]` which I place near the top of a file which will generate the table of contents.

We also have header id attributes which allow us to give id's to specific headers so that we can link to them, the benefit being that we can reference sections within the same file (and I believe across files).

In doxygen we can also create links to things that github has no idea about such as classes, method names and so on, which github would not be able to pick up on.

A relevant link to doxygens way of handling markdown is here: https://www.doxygen.nl/manual/markdown.html#md_header_id

# Option 1

From what I can tell, we could attempt to keep one source for both the github docs and doxygen docs. This means we would have to never use any feature supported by one and not the other. This seems ok for simple markdown files, but contributors who don't know about this set-up may get confused if they attempt to add documentation.

This method would make the use of doxygen for custom `.md` files somewhat irrelevant because it would only ever be able to do what github documentation could do.

# Option 2

We write keep one source for both github and doxygen docs, but we don't care about keeping things compatible with eachother, meaning that there may be dead links, doxygen commands and artifacts like that in the github documentation.

This option makes it easy to contribute, but it lowers the level of quality as seen on the github side of things.

# Option 3

We partition our documentation into files only for github viewing and files only for doxygen viewing.  This means that no documentation file is ever shared between the two and there are no duplicate files.

For doxygen related files, we don't have to have to care to make them compatible with github markdown and we can use any special features.

This option makes contributing more complicated because you have to know if the file you're editing is for github  or doxygen.

# Option 4

Only use doxygen, we keep a basic front page for the github readme, which displays whatever is essential about the project, and link the user directly to doxygen after that. This would keep pretty much all documentation in doxygen, this has the benefit of users not having to think where a markdown file is going to be viewed from, it'll always be doxygen so you don't have to think too hard.

This option makes contributing simple and has the least complexity of any of the other options.

What are your thoughts?

jrouwe:
I've merged what I think is a reasonable compromise. Architecture.md was already doxygen only so it can have a [TOC], the other md files should remain readable from github so I left them as is.

One thing I'm wondering, with the linking into sections is it possible to give sections a link name? Now they're named like:

```
https://jrouwe.github.io/JoltPhysics/#autotoc_md76
```

But if I ever mention a section in a tracker and then add a section before autotoc_md76 the link will be broken.

cuppajoeman:
> I've merged what I think is a reasonable compromise. Architecture.md was already doxygen only so it can have a [TOC], the other md files should remain readable from github so I left them as is.
> 
> One thing I'm wondering, with the linking into sections is it possible to give sections a link name? Now they're named like:
> 
> ```
> https://jrouwe.github.io/JoltPhysics/#autotoc_md76
> ```
> 
> But if I ever mention a section in a tracker and then add a section before autotoc_md76 the link will be broken.
> 
> (Note: I only really care about Architecture.md for this, I don't really link to the other files from issues)

Hey,

What I've been doing in that situation is that I give the section a [header id attribute](https://www.doxygen.nl/manual/markdown.html#md_header_id) so if we had something like

```
## Collision-Filtering
```

We'd augment that to
```
## Collision Filtering {#collision-filtering}
```

And now within that same file we can link via `\ref collision-filtering ...` or using the markdown linking format.

The literal URL of that section on my system when I build looks like this: `file:///home/ccn/Projects/JoltPhysics/Build/Doxygen/md__docs_2_architecture.html#collision-filtering`.

jrouwe:
Ok, I'll do a pass.

jrouwe:
Done, is much nicer like this!

PR #844, opened 2024-01-02
Fixed integer literal type in SetThreadName

mihe:
As of #823 my 32-bit Linux builds are failing in `SetThreadName` with both GCC and Clang due to `error: no matching function for call to 'min'`, which if I'm reading the errors correctly seems to be because `sizeof` is returning an `unsigned int`, which differs from the `unsigned long` literal passed in as the second argument, causing the type deduction to fail as both parameters to `min` are expected to be of the same type.

This changes the `min` call in question to explicitly specify `size_t` as its type to fix this.


jrouwe:
Thanks!

PR #848, opened 2024-01-02
Moving -mfpmath option to Jolt.cmake

jrouwe:


mihe:
I don't quite understand why this latest change runs successfully, as CMake does not (as far as I can tell) have a unary `!` operator.

You might want to change this line:

https://github.com/jrouwe/JoltPhysics/blob/e39aa19c2716810b27ac6c99349f03592abc2b99/Jolt/Jolt.cmake#L556

to be `if (MSVC)` instead.

jrouwe:
I tried that variant as well, unfortunately that doesn't work since MSVC is also true when compiling with clang-cl. This means that the MSVC clang build will go through:

https://github.com/jrouwe/JoltPhysics/blob/e39aa19c2716810b27ac6c99349f03592abc2b99/Jolt/Jolt.cmake#L558-L564

which means it will not add any ```/arch``` flag when compiling with SSE 4.2 or lower. The MS compiler doesn't have the ability to distinguish between the various SSE versions and clang will trigger all sorts of warnings about instructions that it doesn't know (because they haven't been specifically enabled with ```-mXXX```).

PR #859, opened 2024-01-05
Move method for perspective projection matrix

marv7000:
`Mat44` already contains a ton of methods for creating matrices useful for 3D rendering, such as `Mat44::sLookAt`.
My proposal is to move the method for creating perspective projection matrices from `TestFramework/Renderer/Renderer.cpp`, as it makes more sense to have it available as part of the core library.
Maybe an orthogonal projection matrix method would be useful as well.


jrouwe:
Thanks! Initially I put this method in Renderer.cpp because it didn't have anything to do with physics, but if people find it useful I guess it is ok to move it here.

marv7000:
Thanks a ton, appreciate all your work!

PR #884, opened 2024-01-22
Fix BodySetIslandIndex SolvePositionConstraints race condition

ocohen:
I noticed that sometimes bodies would end up with an island even though they were put to sleep. This happens because SolvePositionConstraints puts the body to sleep at the same time BodySetIslandIndex runs. Maybe there's a better fix than how I did it?


jrouwe:
Thanks, I had hoped that no-one would use the GetIslandIndexInternal function (since it's marked very clearly as an 'internal' function and it's only used for debug drawing), but it's indeed a race condition. The job that sets the index is tiny, so it's ok if it needs to finish a little bit earlier.

ocohen:
Just FYI the thing that made me investigate this is that sometimes when debug drawing in island mode an object I knew was inactive was still rendering as if it was active and part of an island.

Thanks for making this library public it has been a pleasure to use

jrouwe:
Thanks for debugging this!

PR #888, opened 2024-01-26
Bugfix: Cast sphere vs triangle could return false positive

### LINKED ISSUE #886: Collision with degenerate line triangle
Hello,
sphere cast with this semi (?)-degenerate triangle causing false-positive (ghost) collision:
```
v[] = {{14.5536213, 10.5973721, -0.00600051880}, {14.5536213, 10.5969315, -3.18638134}, {14.5536213, 10.5969315, -5.18637228}};
```
with this `shapeCast` parameters
```
mStart = {14.8314590, 8.19055080, -4.30825043},
mDirection = {-0.0988006592, 5.96046448e-08, 0.000732421875}
mRadius = 0.2
```
I was wondering if this a programming bug and/or edge case that can be workarounded OR one should filter-out/cleanup it's collision geometry (which require quite a bit of effort, e.g. it's passes area filter/threshold, thats why I very much prefer programming solution)

See also attached obj for real-world usage of said geometry.
[jolt_broken_line_triangle.obj.zip](https://github.com/jrouwe/JoltPhysics/files/14055924/jolt_broken_line_triangle.obj.zip)
image

jrouwe:
Fixes #886

### DIFF
--- a/Jolt/Geometry/ClosestPoint.h
+++ b/Jolt/Geometry/ClosestPoint.h
@@ -14,7 +14,8 @@ namespace ClosestPoint
 {
 	/// Compute barycentric coordinates of closest point to origin for infinite line defined by (inA, inB)
 	/// Point can then be computed as inA * outU + inB * outV
-	inline void GetBaryCentricCoordinates(Vec3Arg inA, Vec3Arg inB, float &outU, float &outV)
+	/// Returns false if the points inA, inB do not form a line (are at the same point)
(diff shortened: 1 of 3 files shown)


virt00l:
I did checked this commit on my test-case and it indeed solve my issue described in #886 
Thanks you very much!

jrouwe:
Thanks for checking, and thanks for reporting the issue!

PR #890, opened 2024-01-26
Made math type args (like Vec3Arg) const so that it remains possible to change them to a const reference

jrouwe:


mihe:
Out of curiosity, what exactly prompted this change? I can't see what this change would allow for exactly.

jrouwe:
The reasons I made the Vec3Arg etc. types is to be able to switch between passing by value or by reference at compile time. You can already see that this is done for DVec3 depending if you have AVX or not (AVX will pass it as a register, non-AVX would copy a big struct so is better passed by reference). In the same way, if you run without SSE registers we could redefine Vec3Arg = const Vec3 & so that the objects are passed by reference. There's no immediate need for this now, but I noticed that some code crept in that modified the in parameters.

mihe:
> There's no immediate need for this now, but I noticed that some code crept in that modified the in parameters.

Ah, right, that makes sense. Thanks for indulging me. :)

PR #894, opened 2024-02-02
Introduce the DEBUG_RENDERER setting

AndreaCatania:
Introduces the possibility to enable the debug renderer even when compiling Jolt in distribution.


jrouwe:
Hello,

Can I ask what the use case for this is? There is quite a bit of code inside the physics engine (besides the official draw calls like PhysicsSystem::DrawBodies) that draws stuff, so I would not recommend enabling the debug renderer in a distribution build as it introduces overhead even if you're not calling draw functions yourself.

What the sample app does in distribution mode is to use ```Shape::GetTrianglesStart``` to display the state of the world. See:

https://github.com/jrouwe/JoltPhysics/blob/cc49e07d7c4eb854a60cb21ea0abdfcc93511410/Samples/SamplesApp.cpp#L2258-L2380

Also, shouldn't there be an `option(DEBUG_RENDERER OFF)` in CMakeLists.txt then?

AndreaCatania:
We are building a game-engine-like App that allows the users to implement their games. For them is good to be able to visualize the physics engine colliders, and the DebugRenderer is the perfect tool for doing that (considering all the built-in functionalities it provides, like the LOD, that we should re-implement if we would use `Shape::GetTrianglesStart`).
The extra overhead is a price we are willing to pay right now, considering that the physics engine is not a bottleneck at the moment.

> Also, shouldn't there be an option(DEBUG_RENDERER OFF) in CMakeLists.txt then?

I'm not a CMake expert, can you please guide me on what to do exactly? Should I add `option(DEBUG_RENDERER OFF)` in:
- https://github.com/AndreaCatania/JoltPhysics/blob/master/Build/Android/UnitTests/src/main/cpp/CMakeLists.txt#L8
- https://github.com/AndreaCatania/JoltPhysics/blob/master/Build/Android/PerformanceTest/src/main/cpp/CMakeLists.txt#L8

jrouwe:
I'll patch it up for you.

AndreaCatania:
@jrouwe Thanks!

Issue #666, opened 2023-08-11
Virtual character has no user data


antopilo:
Hello,

I'm currently using the userData to being able to link back my scene entities to the JoltObjects.

I'm currently using the Character object to simulate my player but I want to be able to use the stepping logic of the VirtualCharacter instead but since the VirtualCharacter object has no UserData, it's making it difficult.

I was just wondering if it was by design that the VirtualCharacter has no UserData.

jrouwe:
The CharacterVirtual class is not inserted into the physics system and the only time you receive callbacks is when you call the (Extended)Update function from your own game object. The callbacks are only triggered on a contact listener which would usually be your game object as well so the context will be clear. I didn't see the need to have user data as you cannot receive random callbacks from random characters, you always know what character you're talking about since you're updating it.

That said, it's easy to add user data, but I'd be curious to know how you structured your code that makes this necessary.

antopilo:
You are right, I was pulling my hair out assuming it was inserted into the physic's system, but in the end I didn't need it. 

Super interesting!

Sorry for the confusion :)

PR #906, opened 2024-02-08
Add UserData to `CharacterVirtual`

skaarj1989:
Even though the `CharacterVirtual` is not inserted into the physics system (#666) I find it useful to have a user data.
+ The size of the `CharacterVirtual` did not change and is equal to 320 bytes.


jrouwe:
Thanks!

PR #913, opened 2024-02-10
Bugfix: Jump button could be missed on high frequency monitors

jrouwe:


mihe:
I believe this bug might be a thing when shooting objects in Samples as well, as nothing happens when I press B sometimes, which is not something I've noticed before.

jrouwe:
Yes, that's indeed the same thing. Fixed now.

PR #916, opened 2024-02-10
Bugfix: Path Constraint: Fixed spelling in enum value

PhoenixIllusion:
I believe this has limited risk of breakage via 'breaking change' to end users, given the recent last spelling change was just this morning.

Searching the project, the other two cases of this spelling are:
`LargeIslandSplitter.(cpp|h)` - private variable, not exposed `mContactAndConstaintsSplitIdx` paralleling `mContactAndConstraintIndices`

`SkeletalAnimation.h` - comment only, `/// Constains the current state of a joint`. I believe the usage should be 'Contain'


jrouwe:
Ok, I can't believe I didn't see this when fixing the enum :)

jrouwe:
Thanks!

PR #925, opened 2024-02-14
Testing

jrouwe:


Build/CMakeLists.txt
```diff
@@ -157,6 +157,11 @@ else()
 		endif()
 	endif()
 
+	# See https://github.com/jrouwe/JoltPhysics/issues/922. When compiling with DOUBLE_PRECISION=YES and CMAKE_OSX_DEPLOYMENT_TARGET=10.12 clang triggers a warning that we silence here.
+	if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
mihe, line 161: I can confirm that this silences the error.

You might want to check the compiler ID instead though, seeing as how this is a Clang-specific flag.

```suggestion
	if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
```

jrouwe: But then it will also be added on iOS, which I don't think we want right? Probably we need both then...

PR #970, opened 2024-03-10
make sure indexing is valid on 4x4 matrices

cuppajoeman:
I was just looking at the source for mat44 and noticed we don't verify the lower bound on indexing, added to the assert to check for that.


godlikepanos:
Isn't that superfluous? The variables you are checking is unsigned and they will always be greater or equal to 0.

jrouwe:
This is indeed redundant code, the variables are unsigned.

cuppajoeman:
Right those are unsigned and already have that constraint, my bad

PR #1019, opened 2024-03-28
Added easy to use implementation of DebugRenderer

jrouwe:
See #1018


eliasdaler:
Thanks! I eventually managed to implement a somewhat nice debug renderer which uses different LODs and does frustum culling, but having a simpler implementation would be very helpful for beginners!

I’ll test this new renderer in my engine tomorrow.

PR #1031, opened 2024-04-03
Update comment for Ragdoll::DriveToPoseUsingMotors()

KyleN97:
Updating comment, may be a better way to describe than what i have used.


jrouwe:
Thanks, that was wrong indeed.

PR #1053, opened 2024-04-13
Fix warnings by commenting out unused parameters.

crocdialer:
warnings occurred when including '**Jolt/Physics/Collision/Shape/MeshShape.h**'. this commit comments out unused parameters of `MeshShape::GetSubmergedVolume`

```
Jolt/Physics/Collision/Shape/MeshShape.h:130:85: error: unused parameter ‘inCenterOfMassTransform’ [-Werror=unused-parameter]
virtual void GetSubmergedVolume(Mat44Arg inCenterOfMassTransform, ...)
```

ubuntu 22.04, gcc 11.4.0

### DIFF
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -308,6 +308,7 @@
 	JPH_GCC_SUPPRESS_WARNING("-Winvalid-offsetof")												\
 	JPH_GCC_SUPPRESS_WARNING("-Wclass-memaccess")												\
 	JPH_GCC_SUPPRESS_WARNING("-Wpedantic")														\
+	JPH_GCC_SUPPRESS_WARNING("-Wunused-parameter")												\
 																								\
 	JPH_MSVC_SUPPRESS_WARNING(4619) /* #pragma warning: there is no warning number 'XXXX' */	\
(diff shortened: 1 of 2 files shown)


jrouwe:
I tried adding -Wextra to CMakelists.txt but the number of errors is just too much. On clang we're also disabling this warning, so I did the same for gcc now.

crocdialer:
nice, can confirm that works for me :)

jrouwe:
Note that Sonar will still find unused parameters, but it's a bit smarter about them.

crocdialer:
yeah, the usage of `JPH_GCC_SUPPRESS_WARNING` does have the downside of unconditionally disabling those warnings for all compilation-units (indirectly) including `Core.h`. and I can imagine how cumbersome it would be to try limiting the scope of those ignores (e.g. wrapping with diagnostic push/pop). anyway, thanks a lot, that was helpful!

jrouwe:
They are wrapped in a push/pop:

https://github.com/jrouwe/JoltPhysics/blob/178141555fe442af3209156bf4b7945ff521bf34/Jolt/Core/Core.h#L362-L365

PR #1060, opened 2024-04-16
Small improvements for e2k

r-a-sattarov:
- added mcst-lcc compiler macro for future use
- added sse macro check before use it

Ref: http://mcst.ru/lcc


jrouwe:
Thanks!

B.t.w. I removed the JPH_COMPILER_MCST_LCC macro (I'd like to avoid having unused macros, we can always put it back if it really needs to be used somewhere)

PR #1066, opened 2024-04-22
Improved handling of mass override for soft bodies

jrouwe:
- If inverse mass override of both the soft body vertex and the colliding body are 0, the collision will be ignored
- If the inverse mass override of the soft body vertex is 0 and the colliding body is kinematic, the collision will be ignored
- If the sensor flag is set, the collision will be reported but not handled


jrouwe:
@mihe I think with this change you can remove your hack:

https://github.com/godot-jolt/godot-jolt/blob/5876f86281a28c5f5071b21f84415489f61d427e/src/spaces/jolt_contact_listener_3d.cpp#L140-L145

mihe:
That does indeed resolve that issue it seems.

Thank you!

PR #1089, opened 2024-05-03
Replace std::vector with custom implementation

jrouwe:


jankrassnigg:
There comes a time when a man has to write their own array class.

Congratulations for this very important achievement :-)

Jolt/Core/Array.h
```diff
@@ -0,0 +1,549 @@
...
+#else
+
+JPH_NAMESPACE_BEGIN
+
+/// Simple replacement for std::vector
+///
+/// Major differences:
+/// - Memory is not initialized to zero (this was causing a lot of page faults)
+/// - Iterators are simple pointers (for now)
+/// - Array capacity can only grow, only shrinks if shrink_to_fit is called
```
SirLynix, line 31: this is actually the same for std::vector, its capacity can never shrink except if shrink_to_fit is called on it.

jrouwe: Thanks! I'll remove that comment.

jrouwe:
> There comes a time when a man has to write their own array class.
> 
> Congratulations for this very important achievement :-)

Yeah, I kept telling myself: std::vector is not that bad and I'll use it until it gets in the way of performance. Unfortunately, I recently learned that std::vector memsets the entire datablock to 0 before doing anything else and there doesn't appear to be a reasonable way around it. This was actually costing quite a bit of time when deserializing big meshes/height fields. And due to my self imposed restrictions that I cannot use 3rd party libs other than STL, I had to write my own 😞

jankrassnigg:
> > There comes a time when a man has to write their own array class.
> > Congratulations for this very important achievement :-)
> 
> Yeah, I kept telling myself: std::vector is not that bad and I'll use it until it gets in the way of performance. Unfortunately, I recently learned that std::vector memsets the entire datablock to 0 before doing anything else and there doesn't appear to be a reasonable way around it. This was actually costing quite a bit of time when deserializing big meshes/height fields. And due to my self imposed restrictions that I cannot use 3rd party libs other than STL, I had to write my own 😞

Yeah, [been there](https://github.com/ezEngine/ezEngine/tree/dev/Code/Engine/Foundation/Containers), [done that](https://github.com/ezEngine/ezEngine/tree/dev/Code/Engine/Foundation/Strings) (for pretty much the entire STL). It's really sad that the STL is barely usable once you actually care about performance.
But well, nothing else one can do about, I guess.

PR #1101, opened 2024-05-10
Declare Job destructor as virtual.

dm-tesla:
This is needed to prevent leaks from user-derived jobs. For example:
```cpp
class JobImpl: final Job {
	const std::string internal_string;
};
```
In this case, the `internal_string` will not be automatically freed.

### DIFF
--- a/Jolt/Core/JobSystem.h
+++ b/Jolt/Core/JobSystem.h
@@ -173,7 +173,7 @@ protected:
 		JPH_OVERRIDE_NEW_DELETE
 
 		/// Constructor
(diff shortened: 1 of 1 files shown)


jrouwe:
Hello,

The `Job` class doesn't have any virtual functions on purpose. There's only 1 way it can be destroyed: `JobSystem::FreeJob` so you can free it like:

```
void MyJobSystem::FreeJob(Job *inJob)
{
	delete static_cast<JobImpl *>(inJob);
}
```

dm-tesla:
I see, thank you for the answer! Closing the issue.
I wish there was a way to guard against that properly. Seeing all those leaks wasn't expected.

PR #1118, opened 2024-05-30
ObjectLayer type miss

Snouf-ID:
When I compile my project on visual, if I am careful not to include too many things, I notice that ObjectLayer mLayer = 0; use an unknown type.


jrouwe:
Thanks!

PR #1129, opened 2024-06-09
Added ScopeExit class

jrouwe:
This class helps to control the lifetime of temp allocations


Jolt/Core/ScopeExit.h
```diff
@@ -0,0 +1,49 @@
...
+
+JPH_NAMESPACE_BEGIN
+
+/// Class that calls a function when it goes out of scope
+template <class F>
+class ScopeExit : public NonCopyable
+{
+public:
+	/// Constructor specifies the exit function
+	JPH_INLINE			ScopeExit(F &&inFunction) : mFunction(inFunction) { }
```
mihe, line 17: I believe you'll want to `std::move(inFunction)` here, to avoid reference collapsing and inadvertently copying the lambda.

PR #1140, opened 2024-06-13
Generate a CMake config file when the project is installed

JimmyCabbage:
I noticed that Jolt only installs its header files and its library file, and that it was lacking a config file for `find_package()` to work automatically.

This commit adds an `export()` and `install(EXPORT)` command in CMakeLists.txt so that one is automatically generated when Jolt is built, and automatically installed when Jolt is installed.

This allows for other CMake projects to do something like:

```CMake
find_package(Jolt)

# snip

add_executable(Generic main.cpp)
target_link_libraries(Generic PRIVATE Jolt::Jolt)
```

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -243,7 +243,9 @@ if (XCODE)
 endif()
 
 # Install Jolt library and includes
-install(TARGETS Jolt DESTINATION lib)
+install(TARGETS Jolt
+	EXPORT JoltExport
+	DESTINATION lib)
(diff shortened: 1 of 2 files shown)


jrouwe:
Thanks, I did not know that this existed!

baszalmstra:
Would you perhaps be able to cut a new release that includes these changes?

jrouwe:
I was planning to create a new release somewhere in the first half of August

jrouwe:
A new version has been released.

baszalmstra:
Much appreciated thanks!

PR #1141, opened 2024-06-13
CMake: Fix SSE not being enabled on x86 32bits

### LINKED ISSUE #1133: -mfpmath=sse is not added in 32bits build
Hi,

-mfpmath=sse is not added in 32bits build, and from a comment in Jolt.cmake this seems to be a mistake

https://github.com/jrouwe/JoltPhysics/blob/4cd52055e09160affcafa557b39520331bf0d034/Jolt/Jolt.cmake#L647

This block is still part of the 64bits check, same for EMIT_X86_INSTRUCTION_SET_DEFINITIONS.

Also, shouldn't -msse2 be added too in 32bits build? -mfpmath=sse is not enough

SirLynix:
Fix #1133


jrouwe:
Thanks for fixing this!

PR #1147, opened 2024-06-30
Resolve mistake in README.md

esuriru:
There is an extra `should` in the last point of the `Design Considerations` section in `README.md`, which should not be there, as it is likely a typo or mistake.


jrouwe:
Thanks!

PR #1149, opened 2024-07-01
We are increasing the CMake required version to fix the configuration issues below.

frederich:
Hi Jorrit,

that's a minor fix to have the correct CMake version in use.

Best,
Jens


jrouwe:
Thank you!

PR #1158, opened 2024-07-04
softbody triangle backface test fix

turanszkij:
When one triangle fails the backface test in RayCast, continue to check other triangles instead of exiting early. This was a problem for me when performing ray cast with `EBackFaceMode::IgnoreBackFaces`, because I was inconsistently missing ray hits.


jrouwe:
That was a silly mistake! Thanks for fixing!

turanszkij:
No problem, I make this kind of mistake all the time. Thanks for making Jolt physics, it's excellent!

PR #1160, opened 2024-07-05
indent_size always works even if it is switched to space

frederich:
Hi Jorrit,

a very very unimportant fix, I discovered it by chance.

By incorrectly editing the wrong .editorconfig (we always use space) I have noticed that the indentation in Jolt no longer works. By setting indent_size it is always given, because indent_size is evaluated in both cases (tab | space).

Best,
Jens

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,11 +3,13 @@ root = true
 [*.{cpp,h,inl,cmake,sh,bat,hlsl}]
 indent_style = tab
 indent_size = 4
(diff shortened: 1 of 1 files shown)


jrouwe:
Thanks, I added tab_width back as I could see it caused problems if the parent .editorconfig contains a different tab_width size.

frederich:
I see, that's right, it's better if it's explicitly stated.

PR #1164, opened 2024-07-06
Allow an external profiler to be installed dynamically

tanner00:
I agree with your point about avoiding the dynamic overhead if it's not necessary, so I ended up adding a check for `JPH_SHARED_LIBRARY`. This may mean that `JPH_SHARED_LIBRARY` needs to be tested for in `RegisterTypesInternal`, but I wasn't sure. Happy to fix anything if necessary :)


jrouwe:
Looks good to me, thanks!

PR #1174, opened 2024-07-13
Added missing JPH_EXPORT macro

jankrassnigg:
The DLL export definition was missing on a few classes, which means you can't derive from them when Jolt is used from a DLL or shared object.


jrouwe:
Thanks!

PR #1175, opened 2024-07-13
Added OldSize argument to realloc function

jankrassnigg:
This is necessary so that implementations that need the old size (for copying the right amount of data over) don't need to track themselves how much memory is associated with the original pointer.


jankrassnigg:
Ah right, I accidentally changed something that didn't need changing, thanks for fixing that.

With this change, I don't need to track the old size on my side anymore, and can remove a hash table and a mutex.

jrouwe:
Thanks again!

PR #1180, opened 2024-07-21
Fixed ScaledShape::MakeScaleValid

mihe:
Currently when passing something like `(0.1, 0.1, 0.1)` into `ScaledShape::MakeScaleValid`, where `ScaledShape::mScale` is `(3, 3, 3)`, you end up getting `(3, 3, 3)` back, whereas you would expect to get `(0.1, 0.1, 0.1)` back.

I assume the divison by `scale` in `ScaledShape::MakeScaleValid` was supposed to be by `mScale` instead, so I went ahead and flipped those.

### DIFF
--- a/Jolt/Physics/Collision/Shape/ScaledShape.cpp
+++ b/Jolt/Physics/Collision/Shape/ScaledShape.cpp
@@ -178,9 +178,9 @@ bool ScaledShape::IsValidScale(Vec3Arg inScale) const
 
 Vec3 ScaledShape::MakeScaleValid(Vec3Arg inScale) const
 {
-	Vec3 scale = ScaleHelpers::MakeNonZeroScale(inScale);
+	Vec3 scale = ScaleHelpers::MakeNonZeroScale(mScale);
 
-	return mInnerShape->MakeScaleValid(mScale * scale) / scale;
+	return mInnerShape->MakeScaleValid(scale * inScale) / scale;
 }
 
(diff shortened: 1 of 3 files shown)


Jolt/Physics/Collision/Shape/ScaledShape.cpp
```diff
@@ -178,9 +178,9 @@ bool ScaledShape::IsValidScale(Vec3Arg inScale) const
 
 Vec3 ScaledShape::MakeScaleValid(Vec3Arg inScale) const
 {
-	Vec3 scale = ScaleHelpers::MakeNonZeroScale(inScale);
+	Vec3 scale = ScaleHelpers::MakeNonZeroScale(mScale);
```
jrouwe, line 181: I think we have to assume that mScale is non zero here. There are other functions that blindly invert the value. Bonus points for adding a check in the constructor and failing shape creation / adding a unit test to verify that this code is correct. (I can do both when I am back from holiday)

mihe: Makes sense. I'll get it done.

mihe: I wasn't sure how crazy I should go with the unit tests. Let me know if I should add some more.

jrouwe: Thanks! I will probably add some extra unit tests later. For now this is fine.

Jolt/Physics/Collision/Shape/ScaledShape.cpp
```diff
@@ -37,6 +37,12 @@ ScaledShape::ScaledShape(const ScaledShapeSettings &inSettings, ShapeResult &out
 	if (outResult.HasError())
 		return;
 
+	if (inSettings.mScale.IsNearZero())
```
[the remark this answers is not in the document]
mihe: Oh right, of course. My bad!

jrouwe:
Thanks! 🙏

PR #1181, opened 2024-07-21
Added overrides for IsValidScale/MakeScaleValid to DecoratedShape

mihe:
This might be by-design perhaps, but I found it a bit odd to have the default implementations of `DecoratedShape::IsValidScale` and `DecoratedShape::MakeScaleValid` not go through its `mInnerShape`, so this adds exactly that.

I removed the now redundant overrides in `OffsetCenterOfMassShape` as well, and changed `RotatedTranslatedShape::IsValidScale` to check with `Shape::IsValidScale` first instead.


jrouwe:
I did this on purpose indeed to make sure I didn't miss any shape, but you're right that it is inconsistent. Thanks

PR #1187, opened 2024-07-30
Enable optional CMake install using ENABLE_INSTALL option

amerkoleci:
Hi,
When using Jolt as static library integrated in other engine would make sense to avoid installation of Jolt.

Added ENABLE_INSTALL  for this purpose.

Amer

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -253,8 +253,7 @@ endif()
 if (ENABLE_INSTALL)
 	install(TARGETS Jolt
 		EXPORT JoltExport
-		DESTINATION lib
-	)
+		DESTINATION lib)
 	foreach(SRC_FILE ${JOLT_PHYSICS_SRC_FILES})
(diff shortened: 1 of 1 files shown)


SirLynix:
Can't something like [this](https://stackoverflow.com/questions/64900981/how-do-you-prevent-cmake-from-install-ing-targets-from-within-projects-include) work without requiring to change Jolt?

amerkoleci:
> Can't something like [this](https://stackoverflow.com/questions/64900981/how-do-you-prevent-cmake-from-install-ing-targets-from-within-projects-include) work without requiring to change Jolt?

Hi, yep, EXCLUDE_FROM_ALL works as well.

jrouwe:
Does this mean this PR can be closed?

amerkoleci:
It doesn't hurt anyway, lot of libs (glfw, sdl), have option to enable/disable install. 

You can choose whether to merge or close.

Cheers

jrouwe:
Ok, I've merged it. Thanks!

PR #1189, opened 2024-08-01
Character: Add GetLayer method

SirLynix:
The character class stores the object layer but the only way to retrieve it is via the body


jrouwe:
Thanks!

PR #1190, opened 2024-08-02
Disable -mfpmath=sse for emscripten builds

guusw:
Hey, first of all thank you for creating this, I'm currently integrating this into our game creation tool and so far it's been great to work with :)

Just a little change I had make to compile the library with emscripten since `-mfpmath` doesn't seem to be a valid compile flag.
Aside from having to set `USE_SSE4_1` externally and enable web simd with `-msimd128 `

### DIFF
--- a/Jolt/Jolt.cmake
+++ b/Jolt/Jolt.cmake
@@ -651,7 +651,7 @@ else()
 			target_compile_options(Jolt PUBLIC -mfma)
 		endif()
 
-		if (NOT MSVC)
+		if (NOT MSVC AND NOT EMSCRIPTEN)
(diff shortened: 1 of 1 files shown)


mihe:
If you update your branch to latest you'll find that this was fixed in 6e0b867a2 already.

You also now have the `USE_WASM_SIMD` CMake option as of 256a61aee, which will add the relevant SIMD flags for you, so you don't have to pass `-msimd128` or set `USE_SSE4_1` anymore.

jrouwe:
@guusw I think mihe is right and this PR can be closed. Let me know if there are any issues still.

Issue #1194, opened 2024-08-03
Core.h defines JPH_USE_NEON


stephengold:
When I compile with g++ version 10 for 64-bit ARM Linux (aarch64), I get numerous errors and warnings:

```console
In file included from /home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Jolt.h:9,

                 from /home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/TriangleSplitter/TriangleSplitterMorton.cpp:5:
/home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Math/UVec4.inl: In member function ‘JPH::UVec4 JPH::UVec4::Swizzle() const’:
/home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Core/ARMNeon.h:85:77: error: there are no arguments to ‘__builtin_shufflevector’ that depend on a template parameter, so a declaration of ‘__builtin_shufflevector’ must be available [-fpermissive]
   85 |  #define JPH_NEON_SHUFFLE_F32x4(vec1, vec2, index1, index2, index3, index4) __builtin_shufflevector(vec1, vec2, index1, index2, index3, index4)
      |                                                                             ^~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Math/UVec4.inl:39:9: note: in expansion of macro ‘JPH_NEON_SHUFFLE_F32x4’
   39 |  return JPH_NEON_SHUFFLE_F32x4(mValue, mValue, SwizzleX, SwizzleY, SwizzleZ, SwizzleW);
      |         ^~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Core/ARMNeon.h:85:77: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
   85 |  #define JPH_NEON_SHUFFLE_F32x4(vec1, vec2, index1, index2, index3, index4) __builtin_shufflevector(vec1, vec2, index1, index2, index3, index4)
      |                                                                             ^~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/jolt-jni/jolt-jni/src/main/native/Jolt/Math/UVec4.inl:39:9: note: in expansion of macro ‘JPH_NEON_SHUFFLE_F32x4’
[...]

PR #1195, opened 2024-08-03
don't assume 64-bit ARM supports NEON

stephengold:
This PR addresses issue #1194.

1. `#define JPH_USE_NEON` is removed Core.h
2. Code to update the FP control word on ARM is selected based on `__aarch64__` and `_M_ARM64` instead of `JPH_USE_NEON`.

I don't fully understand the code being modified, so close review is requested.

For instance, it might be desirable to select `JPH_USE_NEON` via cmake, but I don't know how to do that.

### DIFF
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -172,7 +172,6 @@
 	#define JPH_CPU_ARM
 	#if defined(__aarch64__) || defined(_M_ARM64)
 		#define JPH_CPU_ADDRESS_BITS 64
-		#define JPH_USE_NEON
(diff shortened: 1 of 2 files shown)


stephengold:
I see that broke 5 tests having to do with macOS and determinism. Instead of working by trial and error, I'll leave this to the experts.

jrouwe:
See https://en.wikipedia.org/wiki/ARM_architecture_family#64/32-bit_architecture. Armv8 is the first architecture to support 64 bit and it always supports NEON too, hence that define is defined in Core.h. The issue you're reporting is due to something else and I'll take a look at it.

PR #1207, opened 2024-08-10
Removed redundant /GR trimming for MSVC-like builds

mihe:
With the minimum CMake version having been bumped in #1149 this is no longer needed, as CMake no longer implicitly adds `/GR` to `CMAKE_CXX_FLAGS` as of 3.20 ([CMP0117](https://cmake.org/cmake/help/latest/policy/CMP0117.html)).

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -126,9 +126,6 @@ if (MSVC)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zi")
 	endif()
 
-	# Remove any existing compiler flag that enables RTTI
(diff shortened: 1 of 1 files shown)


jrouwe:
Interesting! Where do you find the time to go through all those cmake release notes? :)

mihe:
Honestly, CMake makes that all pretty digestible, through its [policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html). Any breaking change like this you can opt in or out of as you please.

But, with that said, I've avoided bumping the CMake version myself to avoid any nasty surprises. 😅 I just happened to remember this `/GR` thing when I saw that PR come in, and noted down to make this PR at some point.

PR #1209, opened 2024-08-10
Added options to toggle exception-handling and RTTI

mihe:
I'm not sure how I hadn't spotted this before, but when making #1207 I happened to see that the `/EHsc` flag is implicitly added to Jolt by CMake when compiling with MSVC/clang-cl, and given that you explicitly mention that Jolt doesn't use exceptions in the readme I figured this might be by accident, so I threw this together to see what you'd think about it.

~~The unit tests unsurprisingly all seem to pass still~~, but this does however seem to change the hashes in `PerformanceTest`, which surprised me, so that might need some investigation.

~~`UnitTests` required a bit of song-and-dance on MSVC unfortunately, since doctest's `REQUIRE` macro needs regular C++ exceptions, and re-enabling it for `UnitTests` specifically caused a bunch of `__forceinline` (aka `JPH_INLINE`) to stop being inlined during link-time optimizations, due to mixing exception-handling models, which resulted in link-time warnings that I just couldn't figure out how to disable.~~ Frankly this might be reason enough to put this all behind a CMake option, since this will almost certainly cause issues downstream somewhere.

Anyway, I'm seeing some differences in the binary size of the various Jolt applications (`Distribution`), presumably due to not needing all the unwinding everywhere:

- `HelloWorld.exe`: 2472 KB -> 2440 KB (-11%)
- `JoltViewer.exe`: 2544 KB -> 2499 KB (-2%)
- `PerformanceTest.exe`: 2684 KB -> 2684 KB (0%)
- `Samples.exe`: 4631 KB -> 4502 KB (-3%)
- ~~`UnitTests.exe`: 6701 KB -> 5910 KB (-12%)~~
- Godot Jolt: 9566 KB -> 9540 KB (-0.3%)

~~(Again, `UnitTests` still uses `/EHsc`, so shouldn't see any benefit from this, so those numbers are likely just from replacing `/GL` with `/LTCG`.)~~

The lack of improvement in Godot Jolt is a bit strange, but perhaps I was already seeing the benefits of this due to already disabling exception-handling on my end, while using LTCG/LTO and linking with Jolt statically? I'm just guessing though.

I believe I'm also seeing a slight bump (5%-ish) in `PerformanceTest` when using MSVC, which I guess might be due to something now being able to be inlined, or perhaps improved instruction cache usage, but the laptop I'm testing this on isn't very consistent, so take this with a grain of salt.

I haven't tested this on anything but MSVC yet, so I'll be amazed if this passes CI. I also believe this might break Android builds for me, so I still need to try that.

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -302,18 +302,6 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 			set_property(TARGET UnitTests PROPERTY XCODE_ATTRIBUTE_OTHER_CPLUSPLUSFLAGS[arch=x86_64] "$(inherited) -msse4.2 -mpopcnt")
 		endif()
 
-		# doctest requires exception-handling for its REQUIRE macro, so we re-enable it
-		if (MSVC)
-			target_compile_options(UnitTests PUBLIC /EHsc)
-			# Disable cross-compiland inlining to prevent warnings about __forceinline not being inlined, which is to be expected when mixing exception-handling models
-			set_property(TARGET UnitTests PROPERTY INTERPROCEDURAL_OPTIMIZATION_RELEASE OFF)
-			set_property(TARGET UnitTests PROPERTY INTERPROCEDURAL_OPTIMIZATION_DISTRIBUTION OFF)
-			# Enable link-time code generation explicitly, since it'll be forced to do that anyway, due to the Jolt target using /GL
-			target_link_options(UnitTests PUBLIC /LTCG)
-		else()
-			target_compile_options(UnitTests PUBLIC -fexceptions)
-		endif()
-
 		# Register unit tests as a test so that it can be run with:
 		# ctest --output-on-failure
 		enable_testing()
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -33,6 +33,10 @@ option(INTERPROCEDURAL_OPTIMIZATION "Enable interprocedural optimizations" ON)
 # Note that this currently only works using MSVC. Clang turns Float2 into a SIMD vector sometimes causing floating point exceptions (the option is ignored).
 option(FLOATING_POINT_EXCEPTIONS_ENABLED "Enable floating point exceptions" ON)
 
+# When turning this on, the library will be compiled with C++ exceptions enabled.
+# This adds some overhead and Jolt doesn't use exceptions so by default it is off.
+option(CPP_EXCEPTIONS_ENABLED "Enable C++ exceptions" OFF)
+
 # Number of bits to use in ObjectLayer. Can be 16 or 32.
 option(OBJECT_LAYER_BITS "Number of bits in ObjectLayer" 16)
 
@@ -126,17 +130,17 @@ if (MSVC)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zi")
 	endif()
 
-	# Remove any existing compiler flag that enables exceptions
-	string(REPLACE "/EHsc" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
-
-	# Disable warning about STL and compiler-generated types using noexcept when exceptions are disabled
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4577")
-
 	# Set compiler flag for disabling RTTI
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /GR-")
 
-	if ("${CMAKE_VS_PLATFORM_NAME}" STREQUAL "ARM")
-		# On ARM the exception handling flag is missing which causes warnings
+	if (NOT CPP_EXCEPTIONS_ENABLED)
+		# Remove any existing compiler flag that enables exceptions
+		string(REPLACE "/EHsc" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+
+		# Disable warning about STL and compiler-generated types using noexcept when exceptions are disabled
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4577")
+	else()
+		# Enable exceptions
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc")
 	endif()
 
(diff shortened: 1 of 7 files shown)


jrouwe:
C++ exceptions were indeed supposed to be off (I'm sure at some point they were). The fallout of this change is very unexpected though.

> but this does however seem to change the hashes in PerformanceTest, which surprised me, so that might need some investigation.

The build seems to think that they didn't change. Did you perhaps not define `CROSS_PLATFORM_DETERMINISTIC`?

> Frankly this might be reason enough to put this all behind a CMake option, since this will almost certainly cause issues downstream somewhere.

Yes, agreed.

I'm looking at the std::vector failure now and believe it or not, this is part of the implementation of `shrink_to_fit` for clang:

```C++
#if __cplusplus >= 201103L
  template<typename _Tp, bool
    = __or_<is_copy_constructible<typename _Tp::value_type>,
            is_nothrow_move_constructible<typename _Tp::value_type>>::value>
    struct __shrink_to_fit_aux
    { static bool _S_do_it(_Tp&) noexcept { return false; } };

  template<typename _Tp>
    struct __shrink_to_fit_aux<_Tp, true>
    {
      _GLIBCXX20_CONSTEXPR
      static bool
      _S_do_it(_Tp& __c) noexcept
      {
#if __cpp_exceptions
        try
          {
            _Tp(__make_move_if_noexcept_iterator(__c.begin()),
                __make_move_if_noexcept_iterator(__c.end()),
                __c.get_allocator()).swap(__c);
            return true;
          }
        catch(...)
          { return false; }
#else
        return false;
#endif
      }
    };
#endif
````

In other words, if you disable exceptions then `shrink_to_fit` just becomes a no-op. The standard says:

```
It is a non-binding request to reduce capacity() to size(). It depends on the implementation whether the request is fulfilled.
```

So technically it's not a bug, but it is very surprising that shrinking works in the hard case (with exceptions) and not in the simple case.

**Comment by mihe** [comment]
> The build seems to think that they didn't change. Did you perhaps not define `CROSS_PLATFORM_DETERMINISTIC`?

I did indeed not define it. I just figured the hash would remain the same anyway, but obviously that was naive of me. :) I guess changes in inlining and whatnot is enough to shuffle some bits around.

> this is part of the implementation of `shrink_to_fit` for clang

Technically GCC I guess, since it's from libstdc++ (and not libc++) which Clang uses by default.

> but it is very surprising that shrinking works in the hard case (with exceptions) and not in the simple case.

I had always heard that `shrink_to_fit` was just a best-effort kind of thing, but yes, the fact that they no-op it when exceptions are disabled is odd to say the least.

**Comment by mihe** [comment]
> Technically GCC I guess, since it's from libstdc++ (and not libc++) which Clang uses by default.

Out of curiosity I went to look at [how libc++ does it](https://github.com/llvm/llvm-project/blob/a245a984fd41bc4e9ae225219cb7859d53cd83a4/libcxx/include/vector#L1472-L1490), and they seem to instead just compile out the try-catch, like you would expect, so GCC is the odd one out here I guess.

**Comment by jrouwe** [comment]
> I did indeed not define it. I just figured the hash would remain the same anyway, but obviously that was naive of me. :) I guess changes in inlining and whatnot is enough to shuffle some bits around.

The difference is mostly in 'fast math', if you allow the compiler to freely rearrange expressions then the results differ from compiler to compiler/platform to platform and the hash of the PerformanceTest will differ too.

**Comment by mihe** [comment]
> then the results differ from compiler to compiler/platform to platform

Just to clarify, in this case the hash was different from running `PerformanceTest` without exceptions versus with exceptions, locally on my machine (without `CROSS_PLATFORM_DETERMINISTIC`), with everything else being the same, so I wasn't comparing against the hash stored in `determinism_check.yml` or anything like that.

**Comment by mihe** [comment]
I guess it's also worth mentioning that the `_HAS_EXCEPTIONS=0` definition in MSVC's STL is only "quasi-supported" by them, hence the underscore I guess. I believe it was an internal-ish thing that people just started relying on, so they try to support it as best they can, but it could maybe perhaps cause some odd issues in the future, just so you're aware. I'm also not entirely sure what it does to the ABI of their STL implementation, so I'm not sure whether it necessitates defining it downstream as well when passing STL stuff across the ABI-boundary. I can't imagine that's the case though.

**Comment by jrouwe** [comment]
> Just to clarify, in this case the hash was different from running `PerformanceTest` without exceptions versus with exceptions, locally on my machine (both runs without `CROSS_PLATFORM_DETERMINISTIC`), with everything else being the same, so I wasn't comparing against the hash stored in `determinism_check.yml` or anything like that.

Ok, I assumed indeed you were comparing with determinism_check.yml. In any case, it is ok. With fast math, minor changes to code lead to different hashes. A single function that is inlined differently (and some math that gets rearranged) is enough.

File: Build/CMakeLists.txt [hunk 1]
```diff
@@ -173,6 +183,14 @@ else()
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
 	endif()
 
+	# Set compiler flag for disabling RTTI
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
+
+	# Disable exception-handling
+	if (NOT CPP_EXCEPTIONS_ENABLED)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions")
```
mihe, line 191: It turns out that my Android builds work just fine after all, since I was already disabling exceptions everywhere with the help of `set(CMAKE_ANDROID_EXCEPTIONS FALSE)` in my toolchain file. Although, with that variable being introduced in CMake 3.20 it might just be that the recent CMake version bump in Jolt helped me here.

But, this does result in `-fno-exceptions` being added twice, once from the Android NDK toolchain file and once from this line here. This is fine obviously, and frankly I'm not sure how you'd distinguish between a cross-compilation using the NDK versus the Gradle stuff you're doing anyway. Either way, with this new `CPP_EXCEPTIONS_ENABLED` option it seems unlikely that you'd find yourself with some warning about conflicting/overriding flags that couldn't be resolved by just flipping that option, assuming Clang even emits such a warning in the first place.

jrouwe: Cool. I'm doing a final performance comparison and then I'll merge the change.

jrouwe:
I can confirm that I also see a 5% performance increase on MSVC, there's no measurable difference for clang. I had to turn on cross platform determinism because otherwise the hashes would not be the same for MSVC which makes the results unreliable as we're comparing 2 different simulations.

mihe:
> I can confirm that I also see a 5% performance increase on MSVC

Cool!

> I had to turn on cross platform determinism because otherwise the hashes would not be the same for MSVC which makes the results unreliable as we're comparing 2 different simulations.

Right, I keep forgetting this.

jrouwe:
Thanks for the PR btw!

PR #1210, opened 2024-08-11
Added explicit setting of exception/RTTI flags when enabled

mihe:
This obviously isn't strictly necessary, but I was toying around with trying to mismatch these new options with the values of `CMAKE_ANDROID_EXCEPTIONS` and `CMAKE_ANDROID_RTTI`, and found that even when explicitly enabling something like `CPP_RTTI_ENABLED` while disabling `CMAKE_ANDROID_RTTI` you end up with RTTI disabled in Jolt, which struck me as odd.

Clang doesn't seem to emit any warnings about one flag overriding the other either, which makes this even more precarious, so I figured it's better to just be explicit about it.


jrouwe:
Thanks!

PR #1212, opened 2024-08-11
Workaround for Ninja Multi-Config erroring out

### LINKED ISSUE #1211: Error when generating build files for Ninja and clang-cl
I'm still trying to figure out exactly what's going on, and it's not impossible that this actually is a bug in CMake, but it seems that #1206 broke something.

If you open up Visual Studio's "x64 Native Tools Command Prompt" and run this inside the `Build` folder:

```sh
cmake -S . -B Ninja_ClangCL -G "Ninja Multi-Config" -DCMAKE_MAKE_PROGRAM=C:/Path/To/ninja.exe -DCMAKE_CXX_COMPILER=clang-cl
```

You'll be met with this:

```
-- The CXX compiler identification is Clang 16.0.4 with MSVC-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Path/To/clang-cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done (1.2s)
CMake Error: No output files for WriteBuild! called with comment: Additional output files.
CMake Error: No output files for WriteBuild! called with comment: Additional output files.
-- Generating done (0.5s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Which seems to be caused by this:

https://github.com/jrouwe/JoltPhysics/blob/308dd5779af4da355f9ba9eff8fe8f583b799cc4/Build/CMakeLists.txt#L317

... and this:

https://github.com/jrouwe/JoltPhysics/blob/308dd5779af4da355f9ba9eff8fe8f583b799cc4/Jolt/Jolt.cmake#L504

More specifically the fact that these effectively evaluate to `target_precompile_headers(... PRIVATE "")` for the `ReleaseCoverage` configuration, and CMake doesn't seem to like that empty string there for some reason.

This works just fine with Ninja+MSVC though, so I feel like this is probably a CMake bug, but whatever is to blame it's proving to be a showstopper for me updating Jolt.

EDIT: I just realized that `ReleaseCoverage` configuration isn't present when using MSVC, so I guess that might explain the discrepancy, but things do seem to work fine on Linux with Ninja+Clang, which should have that configuration, so something odd is going on anyway.

jrouwe:
Fixes #1211

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -314,15 +314,26 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 		add_executable(UnitTests ${UNIT_TESTS_SRC_FILES})
 		target_include_directories(UnitTests PUBLIC ${UNIT_TESTS_ROOT})
 		target_link_libraries(UnitTests LINK_PUBLIC Jolt)
-		target_precompile_headers(UnitTests PRIVATE "$<$<NOT:$<CONFIG:ReleaseCoverage>>:${JOLT_PHYSICS_ROOT}/Jolt.h>") # Code coverage doesn't work when using precompiled headers
+
+		# Code coverage doesn't work when using precompiled headers
+		if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config" AND MSVC)
+			# The Ninja Multi-Config generator errors out when selectively disabling precompiled headers for certain configurations.
+			# See: https://github.com/jrouwe/JoltPhysics/issues/1211
+			target_precompile_headers(Jolt PRIVATE "${JOLT_PHYSICS_ROOT}/Jolt.h")
+		else()
+			target_precompile_headers(UnitTests PRIVATE "$<$<NOT:$<CONFIG:ReleaseCoverage>>:${JOLT_PHYSICS_ROOT}/Jolt.h>")
+		endif()
+
 		if (MSVC)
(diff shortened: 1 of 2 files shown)


Build/CMakeLists.txt
```diff
@@ -314,15 +314,26 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 		add_executable(UnitTests ${UNIT_TESTS_SRC_FILES})
 		target_include_directories(UnitTests PUBLIC ${UNIT_TESTS_ROOT})
 		target_link_libraries(UnitTests LINK_PUBLIC Jolt)
-		target_precompile_headers(UnitTests PRIVATE "$<$<NOT:$<CONFIG:ReleaseCoverage>>:${JOLT_PHYSICS_ROOT}/Jolt.h>") # Code coverage doesn't work when using precompiled headers
+
+		# Code coverage doesn't work when using precompiled headers
+		if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config" AND MSVC)
+			# The Ninja Multi-Config generator errors out when selectively disabling precompiled headers for certain configurations.
+			# See: https://github.com/jrouwe/JoltPhysics/issues/1211
+			target_precompile_headers(Jolt PRIVATE "${JOLT_PHYSICS_ROOT}/Jolt.h")
```
mihe, line 322: This should be `target_precompile_headers(UnitTests ...`

jrouwe: oops

PR #1214, opened 2024-08-12
PhysicsLock: Export sLocks

SirLynix:
Fixes compilation when using Jolt as a DLL with asserts enabled


SirLynix:
nvm, it was caused by a config issue on my side.

PR #1222, opened 2024-08-17
Added PlaneShape

jrouwe:


Jolt/Physics/Collision/Shape/PlaneShape.h
```diff
@@ -0,0 +1,143 @@
...
+	// See: ShapeSettings
+	virtual ShapeResult				Create() const override;
+
+	Plane							mPlane;														///< Plane that describes the shape. The negative half space is considered solid.
+
+	RefConst<PhysicsMaterial>		mMaterial;													///< Surface material of the plane
+
+	static constexpr float			cDefaultSize = 1000.0f;										///< Default size of the plane
+
+	float							mSize = cDefaultSize;										///< The bounding box of this plane will run from [-size, size]. Keep this as low as possible for better broad phase performance.
```
mihe, line 36: Somewhat nitpicky, and I realize this is clarified in the comment here, but given that `BoxShape` calls its bounds `half extent` I can't help but feel like maybe this should be referred to as "half size" rather than "size".

The fact that the `GetSize()` says that it's the "size of the bounding box" adds to the confusion a bit as well, since it's technically half the size of the bounding box (along the two perpendicular vectors).

jrouwe: You're right. I've renamed mSize to mHalfExtent.

W.r.t. AABox, it would have been better to name that GetHalfExtent too, but I'm trying to keep the API stable so I'll leave that as is.

PR #1225, opened 2024-08-22
Added ability to store per triangle user data in mesh shape

jrouwe:
* Use MeshShapeSettings::mPerTriangleUserData at about 25% memory increase to get per triangle user data through MeshShape::GetTriangleUserData
* Added Shape::GetLeafShape function to be able to get a leaf shape given a sub shape ID
* Added example and unit test


mihe:
I was just about leave that comment. 😅

PR #1232, opened 2024-08-25
Update project icon/logo

Kruithne:
Hi there,

I have no idea if this is something that's interested in for the scope of the project, but we're using Jolt Physics in our engine, and were interested in having it represented on the splash screen with a modern logo, similar to how other physics libraries, such as Havok, feature.

I'm not a graphics designer, but I noodled something together to bounce the idea around, since there's no point inventing things for someone else to do.

image

image

image

This is the "full" version. I personally think it looks better without the "PHYSICS" but it is part of the project name, so it's up for debate if that should be included. This PR includes a few versions, including the standalone icon, the logo with/without the physics, and just the JOLT by itself. The icon is based on the current docs image.

All assets are included as vector graphics, so they can be scaled and coloured as anyone sees fit. The orange I picked out at random with no specific intention.

image

image

Open to all feedback (or to be told to go away). Submitting PR with the intention that the logo becomes property of the project and it's contributors. Everything used in creation of the logo is CC without attribution for commercial use. Fonts used are Roboto and Facon.


jrouwe:
Hello,

Thank you for putting in the effort of making these logos. Also thanks for wanting to give credits by including a logo!

There are already a couple of logos in this repository that you can use for splash screens:

https://github.com/jrouwe/JoltPhysics/blob/master/Docs/Logo.png
https://github.com/jrouwe/JoltPhysics/blob/master/Docs/LogoDark.png

I realize it's maybe not as modern and slick as the ones you made, but they were drawn by my daughter so it's hard to compete with that :). I may at some point in the future switch to another set of logos, but for now I'd like to keep them the way they are.

PR #1242, opened 2024-09-02
Added Supernova Engine to projects using Jolt

eduardodoria:
Supernva Engine is using Jolt Physics.
https://github.com/supernovaengine/supernova

Live samples:
https://samples.supernovaengine.org/physics3d/
https://samples.supernovaengine.org/physics3d-terrain/


jrouwe:
Thanks!

PR #1264, opened 2024-09-17
Enable upcasting through Ref and RefConst

zopsicle:
Previously, you had to call GetPtr, upcast through the raw pointer, then create a new JPH::Ref. Now you can directly convert JPH::Ref\<T2> to JPH::Ref\<T> if T2 derives from T.


jrouwe:
It doesn't seem to compile...

Note that there is also StaticCast<...>(...) which can up/down cast Ref/RefConst to other types

zopsicle:
It seems to not compile on MSVC, and segfaults in a test on other compilers. Maybe there constructor candidate selection is too fragile after adding new candidates. I’ll use StaticCast, thanks!

PR #1276, opened 2024-09-22
SoftBodyManifold now returns sensor contacts separately

### LINKED ISSUE #1016: Support multiple collisions per vertex for a soft body
See discussion at #936

Right now, there is a limit of a single colliding body per soft body vertex. If the closest body happens to be a sensor this effectively disables the collision with the world and causes artifacts.

We want to be able to track multiple collisions per vertex and we want to be able to report those contacts to client code.

jrouwe:
Before this change, there wasa limit of a single colliding body per soft body vertex. If the closest body happened to be a sensor this effectively disabled the collision with the world and caused artifacts. We can now also detect multiple sensor contacts per soft body.

Fixes #1016


mihe:
Any quirks or gotchas with this? It sounded like it was challenging to come up with a performant implementation for it.

jrouwe:
Well the implementation is not as 'parallel' as I'd like it to be, but I think I found something that doesn't perform too badly.

It differs from godot in the sense that it does vertex level collision detection between the soft body and the sensor, where godot only seems to check if the aabb's overlap (I can make a mode for that if needed).

I'm currently trying to hook it up in godot-jolt to see if I left anything out of the API (I'm suspecting you may need a contact point/normal which I'm currently not exposing).

mihe:
> Well the implementation is not as 'parallel' as I'd like it to be, but I think I found something that doesn't perform too badly.

Cool!

> It differs from godot in the sense that it does vertex level collision detection between the soft body and the sensor, where godot only seems to check if the aabb's overlap (I can make a mode for that if needed).

I struggle to see the AABB mode being a desirable behavior, but then I've said that about other things that later have turned out to be a problem.

> I'm currently trying to hook it up in godot-jolt to see if I left anything out of the API (I'm suspecting you may need a contact point/normal which I'm currently not exposing).

I'd be happy to give this a try myself, if you want. The code surrounding all the area contacts/overlap stuff is a bit rough. I'm not sure if it'll cleanly interface with soft bodies at the moment.

jrouwe:
> I'd be happy to give this a try myself, if you want. The code surrounding all the area contacts/overlap stuff is a bit rough. I'm not sure if it'll cleanly interface with soft bodies at the moment.

That may be a good idea. I got to the point where the area triggers an enter callback for the soft body (but no exit callback), so I think the Jolt API is working. The main problem seems to be that a body inherits from JoltShapedObjectImpl3D while a soft body inherits from JoltObjectImpl3D. There are various places where the code silently returns because it fails to cast to a JoltShapedObjectImpl3D and some of the area implementation is implemented at JoltBodyImpl3D level.

My test scene: [soft-body-cloth.zip](https://github.com/user-attachments/files/17090986/soft-body-cloth.zip)

My branch: https://github.com/jrouwe/godot-jolt/tree/area_vs_soft_body

PR #1290, opened 2024-10-01
Use GNUInstallDirs for installation locations

baszalmstra:
Im working on [adding this fantastic library to conda-forge](https://github.com/conda-forge/jolt-physics-feedstock/). Its yet another package repository. :)

I ran into a small issue where not all files seem to be installed in expected locations. For instance `dll` files should be added to `bin` instead of `lib`. This PR modifies the CMake installation logic to use the locations reported by [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).


jrouwe:
Thanks, that looks like a good change!

PR #1316, opened 2024-11-02
Math:  correct 4 copy-paste errors in header comments

stephengold:
Correct some comments in which subtraction functions were described as addition.


jrouwe:
Thanks!

PR #1330, opened 2024-11-09
PoweredSliderConstraintTest.cpp:  remove unnecessary copy of settings

stephengold:
There's an unnecessary copy of a `BodyCreationSettings` in the sample test for the powered slider constraints.
Eliminating the should copy make the code clearer.
The behavior should be unaffected by this change.


jrouwe:
Thanks, that's a silly mistake!

PR #1349, opened 2024-11-21
Treat AABBTree memory sizes with size_t

dm-tesla:
It's easy to hit the overlow of `uint32_t` when dealing with memory sizes

### DIFF
--- a/Jolt/AABBTree/AABBTreeToBuffer.h
+++ b/Jolt/AABBTree/AABBTreeToBuffer.h
@@ -39,8 +39,8 @@ public:
 		// Estimate the amount of memory required
 		uint tri_count = inRoot->GetTriangleCountInTree(inNodes);
 		uint node_count = inRoot->GetNodeCount(inNodes);
-		uint nodes_size = node_ctx.GetPessimisticMemoryEstimate(node_count);
-		uint total_size = HeaderSize + TriangleHeaderSize + nodes_size + tri_ctx.GetPessimisticMemoryEstimate(tri_count, inStoreUserData);
+		size_t nodes_size = node_ctx.GetPessimisticMemoryEstimate(node_count);
+		size_t total_size = HeaderSize + TriangleHeaderSize + nodes_size + tri_ctx.GetPessimisticMemoryEstimate(tri_count, inStoreUserData);
 		mTree.reserve(total_size);
 
 		// Reset counters
--- a/Jolt/AABBTree/NodeCodec/NodeCodecQuadTreeHalfFloat.h
+++ b/Jolt/AABBTree/NodeCodec/NodeCodecQuadTreeHalfFloat.h
@@ -62,7 +62,7 @@ public:
 	{
 	public:
 		/// Get an upper bound on the amount of bytes needed for a node tree with inNodeCount nodes
-		uint							GetPessimisticMemoryEstimate(uint inNodeCount) const
+		size_t							GetPessimisticMemoryEstimate(uint inNodeCount) const
 		{
 			return inNodeCount * (sizeof(Node) + Alignment - 1);
 		}
--- a/Jolt/AABBTree/TriangleCodec/TriangleCodecIndexed8BitPackSOA4Flags.h
+++ b/Jolt/AABBTree/TriangleCodec/TriangleCodecIndexed8BitPackSOA4Flags.h
@@ -141,7 +141,7 @@ public:
 		}
 
 		/// Get an upper bound on the amount of bytes needed to store inTriangleCount triangles
-		uint						GetPessimisticMemoryEstimate(uint inTriangleCount, bool inStoreUserData) const
+		size_t						GetPessimisticMemoryEstimate(uint inTriangleCount, bool inStoreUserData) const
 		{
 			// Worst case each triangle is alone in a block, none of the vertices are shared and we need to add 3 bytes to align the vertices
 			return inTriangleCount * (sizeof(TriangleBlockHeader) + sizeof(TriangleBlock) + (inStoreUserData? sizeof(uint32) : 0) + 3 * sizeof(VertexData)) + 3;


dm-tesla:
It looks like there are more assumptions about 32-bit address space in the code?
```cpp
uint							mTriangleStart = uint(-1);					// Start of the triangle data in mTree
```
Looking further into this

jrouwe:
Jolt can compile on 32-bit CPUs too, size_t will be 32 bit there, so effectively this change is a no-op for those platforms.

In actual fact, the MeshShape will start to malfunction way before reaching 32 bit numbers because we need to be able to address a triangle in the tree through a `SubShapeID`. If this happens, [this](https://github.com/jrouwe/JoltPhysics/blob/d6c0e27daa1bc5d3b5a56584dc84e7c249bc8c37/Jolt/AABBTree/NodeCodec/NodeCodecQuadTreeHalfFloat.h#L154) error will trigger during construction.

If you need huge triangle meshes, then split them up in blocks and create a body per block.

dm-tesla:
> Jolt can compile on 32-bit CPUs too, size_t will be 32 bit there, so effectively this change is a no-op for those platforms.

Right, and that's totally cool. Using the `size_t` is a great boon to expressiveness here, communicating our intent that this is as big as the memory address, not just a random integer.

> In actual fact, the MeshShape will start to malfunction way before reaching 32 bit numbers because we need to be able to address a triangle in the tree through a SubShapeID. If this happens, [this](https://github.com/jrouwe/JoltPhysics/blob/d6c0e27daa1bc5d3b5a56584dc84e7c249bc8c37/Jolt/AABBTree/NodeCodec/NodeCodecQuadTreeHalfFloat.h#L154) error will trigger during construction.

This is not actually happening. Instead, the `total_size` is computed incorrectly, the allocation is wrong, and we end up with SIGSEGV somewhere here - https://github.com/jrouwe/JoltPhysics/blob/d6c0e27daa1bc5d3b5a56584dc84e7c249bc8c37/Jolt/AABBTree/TriangleCodec/TriangleCodecIndexed8BitPackSOA4Flags.h#L273

Would be good to have a test for this if you don't have already, since it is quite UB...

Also, I'm curious about the reason for `OFFSET_MASK`. Is this strictly to allow using 32-bit addresses? Are you concerned with performance implications of switching it to `size_t` (in which case there will be no need for `OFFSET_MASK`)?

We can indeed address this on our side if we know exactly the conditions that Jolt would treat out of bounds.

jrouwe:
> This is not actually happening. Instead, the total_size is computed incorrectly, the allocation is wrong, and we end up with SIGSEGV somewhere here -

Ok, I will make sure it doesn't do that.

> Would be good to have a test for this if you don't have already, since it is quite UB...

This may actually be prohibitively slow to run as part of a unit test as it would need to construct a mesh of several GBs.

> Also, I'm curious about the reason for OFFSET_MASK. Is this strictly to allow using 32-bit addresses?

The MeshShape structure is a bit packed structure in order to make our game fit into memory, so it stores the number of triangles per node (TRIANGLE_COUNT_BITS = 4) and that leaves (OFFSET_BITS = 28) for encoding the offset to the next block. Naturally this creates a limit to how many triangles can fit in a MeshShape. Unfortunately I cannot put a hard number to that limit as it depends on the tree structure that was built around the mesh.

B.t.w. I see that I changed the MeshShape to use max 8 triangles in the leaf nodes, so I think I can make TRIANGLE_COUNT_BITS = 3 and OFFSET_BITS = 29 now, which would allow meshes 2x as big as the current limit.

dm-tesla:
> This may actually be prohibitively slow to run as part of a unit test as it would need to construct a mesh of several GBs.

Interesting point! I wonder if the test could actually *not* construct several GBs. I.e. All it needs is to ask Jolt to construct a shape and given it a memory region. This memory could be allocated but non-resident (since nothing has to touch it in the test), so it wouldn't impede actual memory use by other tests. Then, ideally, the "mesh is too big" is detected before Jolt needs to actually process the data.

> leaves (OFFSET_BITS = 28) for encoding the offset to the next block

Would it be possible to have these offsets limited, while the total size of `mTree` be bigger? Theoretically, blocks don't have to be placed all in random directions. If they are more or less consecutive, or at least having some order, the offset to the next node would be limited.

> B.t.w. I see that I changed the MeshShape to use max 8 triangles in the leaf nodes, so I think I can make TRIANGLE_COUNT_BITS = 3 and OFFSET_BITS = 29 now, which would allow meshes 2x as big as the current limit.

That would also help (in addition to fixing the crash), thank you!

dm-tesla:
So it sounds like you are not currently interested in supporting non-uint32 addresses. Going to close the PR then.
I do highly recommend still defining a separate type for storing the offsets, so that it's easier to read the code and potentially switch to `size_t` later.

jrouwe:
@dm-tesla I spent some time trying to sort out the issues with very large meshes and extending the number of triangles it can handle. In my test I can now run up to 110M triangles in a single mesh. I also made a test case for it (run `PerformanceTest -s=LargeMesh`), but I'm not running it automatically for now since it uses a lot of RAM and takes several minutes to complete. This is the output it currently produces:

```
N, Num Triangles, Mesh Size, Size / Triangle, SubShapeID Bits, Time
500, 500000, 7820408, 15.6, 24, 0.259
1000, 2000000, 30534616, 15.3, 26, 1.276
1500, 4500000, 67745848, 15.1, 27, 2.887
2000, 8000000, 118666452, 14.8, 28, 5.492
2500, 12500000, 194155504, 15.5, 29, 9.067
3000, 18000000, 267867892, 14.9, 29, 13.198
3500, 24500000, 341143912, 13.9, 29, 19.203
4000, 32000000, 452177300, 14.1, 30, 25.333
4500, 40500000, 619796496, 15.3, 30, 35.517
5000, 50000000, 790591912, 15.8, 31, 44.819
5500, 60500000, 942215028, 15.6, 31, 53.783
6000, 72000000, 1082415336, 15.0, 31, 65.147
6500, 84500000, 1203005544, 14.2, 31, 83.590
7000, 98000000, 1336775284, 13.6, 31, 95.313
7500, 112500000, 1527312072, 13.6, 31, 109.504
```

The highest attempt before failing is for a 7500x7500 square grid that forms a sloping terrain. It takes 1.5 GB of RAM.

> All it needs is to ask Jolt to construct a shape and given it a memory region. This memory could be allocated but non-resident (since nothing has to touch it in the test), so it wouldn't impede actual memory use by other tests. Then, ideally, the "mesh is too big" is detected before Jolt needs to actually process the data.

I don't think this is possible with the current implementation as the size of a MeshShape is very dependent on the actual vertex data.

Can you try out this new version with your data and let me know if this works for you? Also I'm curious how many triangles you're trying to push into a single MeshShape.

jrouwe:
> Theoretically, blocks don't have to be placed all in random directions. If they are more or less consecutive, or at least having some order, the offset to the next node would be limited.

Forgot to respond to this: The triangle vertices are using this concept. However, the `SubShapeID` needs to be able to address a particular triangle in the shape, which means it can only be defined relative to the tree root. It is a 32-bit number because that nicely packs in some of the data structures. In our games, meshes are rarely bigger than 10MB, so 32-bit is more than enough.

PR #1350, opened 2024-11-22
Removed the use of `mt19937`

jrouwe:
It started to generate different random numbers between Linux/Windows

### DIFF
--- a/.github/workflows/determinism_check.yml
+++ b/.github/workflows/determinism_check.yml
@@ -2,7 +2,7 @@ name: Determinism Check
 
 env:
   CONVEX_VS_MESH_HASH: '0x16ca5bf7f9da5f74'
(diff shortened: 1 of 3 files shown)


jrouwe:
The issue in this case was the uniform_real_distribution, see the commit msg:

> Removed the use of std::uniform_real_distribution (https://github.com/jrouwe/JoltPhysics/pull/1350)
When going from MSVC 19.41.34123.0 -> 19.42.34433.0 it started to generate different numbers between Windows and Linux, causing the cross platform determinism check to start failing. On Linux, std::uniform_real_distribution is implemented using doubles and on Windows it is not. Removed the use of random from this test.

I initially thought it was the random generator that was different, hence the title of the PR.

PR #1370, opened 2024-12-02
CharacterBaseTest:  delete 2 useless assignments

stephengold:
I noticed 2 assignments in CharacterBaseTest loops that appear to have no effect, since the lvalue is dead.
Perhaps these were the result of copying and pasting the assignment from line 436.


jrouwe:
Yes, that's indeed what happened👍

PR #1372, opened 2024-12-04
Add serialization to SkeletalAnimation

GustavoSilvera:
Small changes to add `SkeletalAnimation::SaveBinaryState` and `SkeletalAnimation::sRestoreFromBinaryState` features for serializing and deserializing the `SkeletalAnimation` instances.

Also added a public facing API to change the `mIsLooping` flag which was previously always `true`


jrouwe:
Thanks!

PR #1376, opened 2024-12-05
Samples/Tests:  correct animation filenames

stephengold:
The sample tests contain hard-coded filenames for animations, such as "Assets/Human/Dead_Pose1.tof".

On case-sensitive filesystems, the coded names do not match the actual filenames in the repository, in this instance "dead_pose1.tof".

Currently, the sample tests only run on Windows, so the mismatches do not cause failure. I discovered the mismatches because I am converting the tests to run on Linux.

### DIFF
--- a/Samples/Tests/General/MultithreadedTest.cpp
+++ b/Samples/Tests/General/MultithreadedTest.cpp
@@ -151,7 +151,7 @@ void MultithreadedTest::RagdollSpawner()
 	{
 #ifdef JPH_OBJECT_STREAM
 		Ref<SkeletalAnimation> animation;
-		if (!ObjectStreamIn::sReadObject("Assets/Human/Dead_Pose1.tof", animation))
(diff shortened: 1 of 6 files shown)


jrouwe:
Thanks!

How are you porting these examples? What renderer are you using?

stephengold:
For now I'm porting the examples without any 3-D rendering or GUI, in order to smoke-test [my JVM bindings](https://github.com/stephengold/jolt-jni). Eventually I plan to render them (from Java) using [LWJGL](https://www.lwjgl.org/), which can bind to either OpenGL rendering or Vulkan rendering.

jrouwe:
Ok, that's good. I'm planning to (one day) write a Vulkan version of the sample app renderer so that the samples can run natively on Linux.

stephengold:
That sounds great. I'll keep an eye out for pitfalls.

jrouwe:
FYI: The samples run natively on Linux now (using Vulkan)

PR #1377, opened 2024-12-05
PowerPC 64bit support

Link4Electronics:
Trying to add PowerPC support, it works as I tested on the project [OpenTESArena](https://github.com/afritz1/OpenTESArena) that uses JoltPhysics
for the TickCounter I was trying to implement like this [code](https://github.com/bvdberg/code/blob/master/ppc/rdtsc/rdtsc.h) but didn't figure it out: 

On Core.h, I don't know why it doesn't try to detect powerpc 32bits or 64bits, I had to comment the if condition to make it work for 64bits


Jolt/Core/Core.h
```diff
@@ -191,6 +191,16 @@
 		#define JPH_USE_SSE4_1
 		#define JPH_USE_SSE4_2
 	#endif
+#elif defined(__ppc64__) || defined(_M_PPC64) || defined(__powerpc__) || defined(_M_PPC)
+//	#if defined(__ppc64__)
+	#define JPH_CPU_PPC64
+		#define JPH_CPU_ADDRESS_BITS 64
```
jrouwe, line 197: We use JPH_CPU_ADDRESS_BITS to distinguish between 32 and 64 bit. So this should be JPH_CPU_PPC in both cases.

Jolt/Core/Core.h
```diff
@@ -191,6 +191,16 @@
...
+/*	#else
+	#define JPH_CPU_PPC32
```
jrouwe, line 199: Please remove commented out code

Jolt/Core/FPControlWord.h
```diff
@@ -126,6 +126,35 @@ class FPControlWord : public NonCopyable
 	uint32		mPrevState;
 };
 
+#elif defined(JPH_CPU_PPC64)
+
+template <uint64 Value, uint64 Mask>
```
jrouwe, line 131: Does this code work? I would expect a change in FPException.h and FPFlushDenormals.h too.

Jolt/Math/Math.h
```diff
@@ -122,6 +122,14 @@ inline uint CountTrailingZeros(uint32 inValue)
 	#endif
 #elif defined(JPH_CPU_E2K)
 		return inValue ? __builtin_ctz(inValue) : 32;
+#elif defined(JPH_CPU_PPC64)
```
jrouwe, line 125: This code is the same as that for JPH_CPU_E2K so it can just be `#elif defined(JPH_CPU_E2K) || defined(JPH_CPU_PPC)`

Jolt/Math/Math.h
```diff
@@ -152,6 +160,14 @@ inline uint CountLeadingZeros(uint32 inValue)
 	#endif
 #elif defined(JPH_CPU_E2K)
 		return inValue ? __builtin_clz(inValue) : 32;
+#elif defined(JPH_CPU_PPC64)
```
jrouwe, line 163: Same remark as above.

Jolt/Core/Core.h
```diff
@@ -191,6 +191,16 @@
 		#define JPH_USE_SSE4_1
 		#define JPH_USE_SSE4_2
 	#endif
+#elif defined(__ppc64__) || defined(_M_PPC64) || defined(__powerpc__) || defined(_M_PPC)
```
jrouwe, line 194: I would expect many more code changes to be needed. As far as I understand PPC is a big endian architecture and Jolt makes quite a few assumptions about it being little endian. I don't really want to have to take endian swapping into account everywhere as it results in really messy code. Did you try running the unit tests? 

I would add a build here: https://github.com/jrouwe/JoltPhysics/blob/master/.github/workflows/build.yml which runs the unit tests using `qemu-powerpc` similar to https://github.com/jrouwe/JoltPhysics/blob/e0e434f59c2ff6816e07af7f6f6fcf9726e2aa0f/.github/workflows/determinism_check.yml#L154-L171

jrouwe (changes requested):
I think the OpenTESArena project is really cool! What hardware are you running this on? I wouldn't know where to get something that runs on PowerPC hardware. Does godot itself even run on that?

Link4Electronics:
> What hardware are you running this on? I wouldn't know where to get something that runs on PowerPC hardware.
> 
> Edit: Sorry, I'm answering too many questions at the same time. Confusing a godot request with this one.



> What hardware are you running this on? I wouldn't know where to get something that runs on PowerPC hardware.
> 
> Edit: Sorry, I'm answering too many questions at the same time. Confusing a godot request with this one.

Sorry it took me a while to answer, I'm running on a PowerMac G5, Debian sid ppc64 and GCC 14. Well, it would help to have a port for the AmigaOne machines like X1000, X5000, sam460ex, sony ps3 and maybe an xbox360 port (all of them are powerpc 64bits), for those using Linux or AmigaOS on those platforms.

About some parts being the same as E2k is according to [this](https://en.wikipedia.org/wiki/Find_first_set) I was trying to follow these examples.

But I forgot to run the unit tests, gonna try them as soon as possible.

Best regards,
Link.

jrouwe:
Hello,

The platforms you mention are all very old. Jolt's design means that it is willing to spend some extra CPU cycles in order to better parallelize things. On these old platforms every CPU cycle counts, so that design decision doesn't really make sense and I'm expecting very poor performance as a result.

I think much bigger changes are going to be needed to deal with endianness problems and then the added maintenance burden for me is just not worth it at this moment. So, it is unlikely that I will accept this PR, sorry.

Link4Electronics:
Hello,

Yes you're right, I understand that.
Just out of curiosity
```test cases: 312 | 304 passed | 8 failed | 172 skipped
assertions: 54896 | 54756 passed | 140 failed |
Status: FAILURE!```

I guess this PR could be closed.

jrouwe:
Turns out that the godot project also wants to build against PPC, so I have reopened the PR and will look at merging it.

octanejohn:
i think is better to focus only on ppc64le (little endian) not ppc64 (big endian) as most linux software is for that

PR #1404, opened 2024-12-19
Update docs to point to SetSimShapeFilter

LPGhatguy:
In the original commit adding the simulation shape filter, the member was called `SetSimulationShapeFilter`, but then it was renamed to `SetSimShapeFilter` in a follow-up. This PR updates the docs to point to the new name.


jrouwe:
Thanks!

Issue #298, opened 2022-10-24
ARM64 support


amerkoleci:
Hi,
I'm trying to add ARM64 support for my C# bindings (Windows and MacOS), but MacOS fails when I compile for ARM64.

Any thought?

Those are the CMake args I used:

https://github.com/amerkoleci/JoltPhysicsSharp/blob/main/build.cake#L32

jrouwe:
It looks like your errors are:

```
clang: error: argument unused during compilation: '-msse4.2' [-Werror,-Wunused-command-line-argument]
clang: error: argument unused during compilation: '-mpopcnt' [-Werror,-Wunused-command-line-argument]
```

And it's probably caused because you're cross compiling from x86 to ARM. I only tested compiling on MacOS ARM for ARM and MacOS x86 for x86. 

The issue is probably in:

https://github.com/jrouwe/JoltPhysics/blob/master/Build/CMakeLists.txt#L151

where it checks if the target platform is x86:

```
elseif ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64")
```

and adds the x86 specific compiler options if it is. Unfortunately the [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) variable is not consistent [according to this](https://stackoverflow.com/questions/11944060/how-to-detect-target-architecture-using-cmake).

I don't own a Mac so testing is a bit hard for me. What you can do is add a:

```
message("CMAKE_SYSTEM_PROCESSOR ${CMAKE_SYSTEM_PROCESSOR}")
message("CMAKE_OSX_ARCHITECTURES ${CMAKE_OSX_ARCHITECTURES}")
```

before that if block to see what the value is set to. If CMAKE_OSX_ARCHITECTURES provides the right platform name on Mac, you could perhaps extend the 'if' so that it takes the ARM branch instead of the x86 branch (and if you do fix it please send a pull request).

amerkoleci:
I think thats the Linux part, the MacOS compiler probably end here:

https://github.com/jrouwe/JoltPhysics/blob/master/Build/CMakeLists.txt#L113

amerkoleci:
Ok,
[...]

PR #1422, opened 2024-12-28
Fix Vulkan descriptor pool count

Wunkolo:
Allocates the correct amount of UBOs and Image Samplers ahead of time.

Thank you for the Vulkan implementation. I did some testing and ran into a crash and validation error regarding the global descriptor-pool here not declaring enough descriptors of certain types ahead of time.

This addresses the issue directly, but a more robust solution might be needed here later on. This is enough to get all of the tests to run without any crashes on my system or validation _errors_ in particular, though I do get an allocation error involving the application reaching the maximum amount of allocations(4096 on this system) when running `FunnelTest` in particular. That will probably need a whole new allocation strategy though to sub-allocate out of a larger heap of VkDeviceMemory rather than allocating a new one for each DebugDraw TriangleBatch.

For reference, I tested on my ThinkPad X13S(Windows on ARM, Native Vulkan driver) and got valdation messages like this followed by a crash due to the global descriptor pool only allocating space for a `cFrameCount`-amount of uniform-buffers despite there being descriptor-layouts that want two uniform-buffers each descriptor set or one singular texture sampler per-set:

```
VK: Validation Error: [ VUID-VkDescriptorSetAllocateInfo-apiVersion-07896 ] Object 0: handle = 0x27d60e0000000019, type = VK_OBJECT_TYPE_DESCRIPTOR_POOL; | MessageID = 0x5584bd53 | vkAllocateDescriptorSets():  Unable to allocate 1 descriptors of type VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER from VkDescriptorPool 0x27d60e0000000019[]. This pool only has 0 descriptors of this type remaining. The Vulkan spec states: If the VK_KHR_maintenance1 extension is not enabled and VkPhysicalDeviceProperties::apiVersion is less than Vulkan 1.1, descriptorPool must have enough free descriptor capacity remaining to allocate the descriptor sets of the specified layouts (https://vulkan.lunarg.com/doc/view/1.3.290.0/mac/1.3-extensions/vkspec.html#VUID-VkDescriptorSetAllocateInfo-apiVersion-07896)
```

### DIFF
--- a/TestFramework/Renderer/VK/RendererVK.cpp
+++ b/TestFramework/Renderer/VK/RendererVK.cpp
@@ -377,13 +377,16 @@ void RendererVK::Initialize()
 	FatalErrorIfFailed(vkCreatePipelineLayout(mDevice, &pipeline_layout, nullptr, &mPipelineLayout));
 
 	// Create descriptor pool
-	VkDescriptorPoolSize descriptor_pool_size = {};
-	descriptor_pool_size.type = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
-	descriptor_pool_size.descriptorCount = cFrameCount;
+	VkDescriptorPoolSize descriptor_pool_sizes[] = {
+		// ubo_layout_binding * (Projection + Ortho)
+		VkDescriptorPoolSize{VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, cFrameCount * (2 * 2)},
+		// texture_layout_binding * (Main Render Targets + DebugUI)
+		VkDescriptorPoolSize{VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, cFrameCount * (1 * 2)},
+	};
 	VkDescriptorPoolCreateInfo descriptor_info = {};
 	descriptor_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO;
-	descriptor_info.poolSizeCount = 1;
-	descriptor_info.pPoolSizes = &descriptor_pool_size;
+	descriptor_info.poolSizeCount = std::size(descriptor_pool_sizes);
+	descriptor_info.pPoolSizes = descriptor_pool_sizes;
 	descriptor_info.maxSets = 256;
 	FatalErrorIfFailed(vkCreateDescriptorPool(mDevice, &descriptor_info, nullptr, &mDescriptorPool));
 


jrouwe:
Hello,

Thanks for looking into this! I'm pretty new to Vulkan and I relied a lot on the validation errors that my particular driver triggered (RTX 3070 Ti).

> I do get an allocation error involving the application reaching the maximum amount of allocations(4096 on this system) when running FunnelTest in particular. That will probably need a whole new allocation strategy though to sub-allocate out of a larger heap of VkDeviceMemory rather than allocating a new one for each DebugDraw TriangleBatch.

Good to know. My driver reports the same max amount of allocations and I thought I got lucky and stayed below that number. Apparently the NVidia driver has a higher internal limit than what it reports. I'll write a simple memory allocator for it.

> I tested on my ThinkPad X13S(Windows on ARM, Native Vulkan driver)

Out of curiosity: Did this not run properly with the DirectX version?

Wunkolo:
> Good to know. My driver reports the same max amount of allocations and I thought I got lucky and stayed below that number. Apparently the NVidia driver has a higher internal limit than what it reports. I'll write a simple memory allocator for it.

I've noticed Nvidia will tend to be very forgiving and "cushiony" involving compliance while AMD/Intel/etc can be more rigid. I got driver-crashes on my ThinkPad X13S for the descriptor-set issue, but did _not_ get one for the allocation-count issue once I dismissed the validation messages and allows the program to continue. It's likely the allocation-crash is more nuanced where it can be the allocation-count _and_ the combined total size of allocations and other nuances that are difficult to advertise to Vulkan.

It took until [VK_KHR_maintenance4](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_maintenance4.html) to get the actual  [maximum size allowed for a _singular_ allocation](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#limits-maxBufferSize).

> Out of curiosity: Did this not run properly with the DirectX version?

DirectX ran just fine! I tested it some time ago back when ARM64 was first brought online and it still works just great!
https://github.com/jrouwe/JoltPhysics/issues/298#issuecomment-1311053580

This just happens to be the laptop I'm traveling with at the moment and is probably the quirkiest hardware I can test on.

jrouwe:
> DirectX ran just fine! I tested it some time ago back when ARM64 was first brought online and it still works just great!
https://github.com/jrouwe/JoltPhysics/issues/298#issuecomment-1311053580

Ah yes, I totally forgot about that :)

jrouwe:
@Wunkolo I've written a crude memory allocator for Vulkan to stay well below the 4096 allocation limit in #1429. It's a bit more wasteful than something like [VMA](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator), but it should be good enough for the samples.

Wunkolo:
Just tested it locally myself and it does drastically reduce the maximum allocation warning! The "Funnel" sample-scene now works just fine for me. I do get "This app has over 250 memory object" validation-warnings on that scene but I generally do not get any crashes anymore when running all of the tests.

PR #1423, opened 2024-12-28
Implement Vulkan transient depth buffer

Wunkolo:
The depth buffer is cleared/uninitialized at the start of a render pass and discarded at the end and does not need to be backed by video memory at all. This depth buffer texture can possibly be backed [transiently](https://docs.vulkan.org/samples/latest/samples/performance/subpasses/README.html#_transient_attachments) rather than allocating unused device memory("memoryless" on MacOS).

This will test and utilize transient memory when available.


jrouwe:
This doesn't seem to work on my GPU. It reports that it supports transient memory but then fails in `RendererVK::FindMemoryType`:

```
		inTypeFilter	3	unsigned int
		inProperties	16	unsigned int
-		mMemoryProperties.memoryTypes	0x0000023fb05376fc {{propertyFlags=0 heapIndex=1 }, {propertyFlags=1 heapIndex=0 }, {propertyFlags=1 ...}, ...}	VkMemoryType[32]
+		[0]	{propertyFlags=0 heapIndex=1 }	VkMemoryType
+		[1]	{propertyFlags=1 heapIndex=0 }	VkMemoryType
+		[2]	{propertyFlags=1 heapIndex=0 }	VkMemoryType
+		[3]	{propertyFlags=6 heapIndex=1 }	VkMemoryType
+		[4]	{propertyFlags=14 heapIndex=1 }	VkMemoryType
+		[5]	{propertyFlags=7 heapIndex=0 }	VkMemoryType
+		[6]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[7]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[8]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[9]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[10]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[11]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[12]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[13]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[14]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[15]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[16]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[17]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[18]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[19]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[20]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[21]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[22]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[23]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[24]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[25]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[26]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[27]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[28]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[29]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[30]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
+		[31]	{propertyFlags=0 heapIndex=0 }	VkMemoryType
```

There's nothing in this list that passes:

```
(mMemoryProperties.memoryTypes[i].propertyFlags & inProperties) == inProperties
```

I'm not sure how to fix this.

jrouwe:
B.t.w. the way that I read the page you mention is that subpasses support transient memory, but render passes (which is what I had to use because the shadow map has different dimensions than the frame buffer) do not.

Wunkolo:
Looks like Nvidia and other non-tiled architectures support transient attachments but does not have a memory that supports `LAZILY_ALLOCATED_BIT` like tile-architecture GPUs do and that particular memory-property has to be tested for and utilized separately!

> I read the page you mention is that subpasses support transient memory, but render passes (which is what I had to use because the shadow map has different dimensions than the frame buffer) do not.

I believe these memory operation nuances exist both in the begin/end of a render-pass and during the transitions between sub-passes. When an image-attachment within one of these passes never actually has to read or write to main video memory, it can be sustained in faster memory without having to do a round-trip between sub-passes or when starting/ending a render-pass. Something like shadow-passes though _do_ write their output into main memory at some point with a `VK_ATTACHMENT_STORE_OP_STORE` at some point in its render pass for later shadow-sampling and would not benefit from this PR though.

This is maybe a bit of a better explanation for it.
https://community.arm.com/arm-community-blogs/b/graphics-gaming-and-vr-blog/posts/picking-the-most-efficient-load-store-operations

The Vulkan API design gets very explicit and tricky here involving render-passes and sub-passes and such.
A framebuffer/renderpass-object may have dozens of image-attachments, but only a subset of [maxColorAttachments](https://registry.khronos.org/vulkan/specs/latest/man/html/VkPhysicalDeviceLimits.html) color-attachments([typically this is 8](https://vulkan.gpuinfo.org/displaydevicelimit.php?name=maxColorAttachments&platform=all)) and a depth/stencil-attachment can actually be addressable within an actual subpass where the rendering actually happens. Because of this, sometimes image-attachments have to be `VK_ATTACHMENT_STORE_OP_STORE`-ed and `VK_ATTACHMENT_LOAD_OP_LOAD`-ed again between passes or preserved and mix-and-matched.

This test-framework is very simple though with render-passes with singular sub-passes for rendering into the swapchain and for the shadow-pass and doesn't need any more complexity involved unless some more intricate rendering techniques get involved. In this case we would be saving up 8MiB of vram at 1080p and a good chunk of memory-bandwidth with this PR by utilizing this optimization(`1920 * 1080 * sizeof(float)` in the case of `VK_FORMAT_D32_SFLOAT`/`VK_FORMAT_D24_UNORM_S8_UINT`) so I figured this might be a worthwhile saving for the hardware that supports it.

jrouwe:
Thanks for explaining and thanks for the PR! (For some reason I thought your code applied to the shadowmap, but it is only the main depth buffer and that one indeed doesn't need preserving)

PR #1425, opened 2024-12-29
Samples and JoltViewer can run on Linux now

jrouwe:


jmiskovic:
A small improvement is needed - the `ui.tga` should be renamed to `UI.tga` so it can be found on case-sensitive filesystems.

Samples are running great. Thank you for this effort!

jrouwe:
Ah, I'm running on WSL, so I didn't notice. Will fix immediately.

PR #1436, opened 2025-01-04
Improved mouse, keyboard, and resize handling for macOS

warrenm:
This PR introduces a few small changes improving the newly added support for macOS in the test framework:

 - `GCKeyboard` operates outside of the Cocoa event loop, and the application window does not handle key events itself, resulting in an annoying beep on every key-down or key repeat event. Introducing a local event monitor and explicitly ignoring key-down events (excluding Command-Q, which is the Quit menu accelerator key) prevents this.

 - Because of the Cocoa view lifecycle, it is possible for a resize event to occur before the window resize handler is installed by the renderer, leading to a crash (on macOS Sequoia 15.2 at least). Adding a null check (i.e. checking whether the callback function is empty) prevents this.

 - Finally, the code currently assumes that window coordinates match backing layer coordinates, meaning the menu doesn't respond correctly to mouse hover or mouse click events. Refining the mouse event handling code to convert into the expected coordinate space restores menu functionality.

### DIFF
--- a/TestFramework/Input/MacOS/KeyboardMacOS.mm
+++ b/TestFramework/Input/MacOS/KeyboardMacOS.mm
@@ -75,6 +75,16 @@ static EKey sToKey(GCKeyCode inValue)
 - (KeyboardDelegate *)init:(KeyboardMacOS *)Keyboard
 {
 	mKeyboard = Keyboard;
+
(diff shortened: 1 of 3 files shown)


jrouwe:
Thanks for fixing this!

PR #1439, opened 2025-01-05
Body.cpp:  correct a typo in a diagnostic message

stephengold:
An extraneous "t" crept into the assertion message in line 42 of Body.cpp . This PR removes it.


jrouwe:
Thanks!

PR #1491, opened 2025-02-03
Support 64-bit WebAssembly

Repiteo:
Now that `MEMORY64` is [no longer experimental](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3172---111924), we'll probably be seeing more emscription builds targeting 64-bit in the coming months. This PR *should* be all that's needed to get it to compile


jrouwe:
Thanks!

PR #1511, opened 2025-02-15
Add LÖVR

bjornbytes:
LÖVR uses Jolt now!  Thank you for creating this library 💜

### DIFF
--- a/Docs/ProjectsUsingJolt.md
+++ b/Docs/ProjectsUsingJolt.md
@@ -7,6 +7,7 @@
 * [Horizon Forbidden West](https://www.playstation.com/en-us/games/horizon-forbidden-west/) - An open world action RPG adventure.
 * [HypeHype](https://www.hypehype.com/) - A mobile app to create, remix and play games. See [this](https://twitter.com/SebAaltonen/status/1726871354228482237) X post.
 * [Light Tracer Render](https://lighttracer.org/) - A rendering tool, uses Jolt for object placement. See [this](https://lighttracer.org/blog/light-tracer-render-2-4-0/) announcement.
+* [LÖVR](https://lovr.org) - A Lua VR framework.  See [this](https://lovr.org/docs/v0.18.0) release announcement.
 * [The Mirror](https://themirror.space/) - A game development platform designed to empower developers and artists with real-time, limitless creativity. See [this](https://twitter.com/themirrorgdp/status/1718019599361323023?s=20) X post.
 * [NeoAxis Engine](https://www.neoaxis.com/) - A 3D game engine. See [this](https://www.neoaxis.com/news/neoaxis_engine_2023_1_released) announcement.
 * [Sceneri](https://www.sceneri.com/) - A mobile app for creating and sharing 3D games and experiences. See [this](https://www.sceneri.com/blog/2023-07-27-jolt-physics-bringing-sceneris-worlds-to-life) blog post.


jrouwe:
Thanks, that's very cool! It has been quite a project I see (about a year)! I'd be interested to know what kind of things you ran into and if there's missing documentation / unintuitive APIs etc.

bjornbytes:
Initially Jolt was a challenge to integrate, since the C bindings were immature.  But we contributed a lot of features to Jolt-Physics-Sharp and eventually it got split off into joltc, which is now very comprehensive and self-contained.  (Thanks to @amerkoleci for merging all my PRs!)

A lot of Jolt features I was missing got added over time: EmptyShape, PlaneShape, triangle index for raycasts.

Probably the hairiest part of Jolt was managing all of the shapes.  Conceptually, LÖVR models physics bodies as having a list of shapes attached to them.  The shapes can each have an offset, and the body can have a custom mass/COM.  Whenever someone adds/removes a shape, or adjusts the mass stuff, we have to do a bunch of gymnastics: promote to a `MutableCompoundShape` if needed, switch to an `EmptyShape` if last shape was removed, wrap in an `OffsetCenterOfMassShape` to preserve their custom center of mass if needed, notify all the joints if COM changed, etc. etc.  There's a bit of a combinatorial explosion of cases that need to be handled to keep things consistent, and I'm still not sure that I got it all right, see [here](https://github.com/bjornbytes/lovr/blob/0ab47ba8f13a06419588d6a4c295ab58f20aea59/src/modules/physics/physics.c#L1029-L1060) for an example in the `lovrColliderAddShape` function.  Part of me wishes that Jolt modeled bodies as having a list of shapes, and that the compound/decorator shapes weren't necessary, but it would be less flexible, make structs bigger, and I imagine this design is necessary due to thread safety.

Also, having a small library size was really important.  I did some measurements and found that the object stream was inflating code size a lot, but then you added `ENABLE_OBJECT_STREAM` a few days later :smile:.  Currently, LÖVR also doesn't use characters, vehicles, ragdolls, or soft body, and it would be nice to have the option of disabling those to save some size.  I wasn't able to figure out how to yank them out cleanly though, and it didn't seem like the savings were going to be as impressive as object stream, so I moved on.

But overall, Jolt has been really great!

amerkoleci:
> Initially Jolt was a challenge to integrate, since the C bindings were immature. But we contributed a lot of features to Jolt-Physics-Sharp and eventually it got split off into joltc, which is now very comprehensive and self-contained. (Thanks to @amerkoleci for merging all my PRs!)
> 
> A lot of Jolt features I was missing got added over time: EmptyShape, PlaneShape, triangle index for raycasts.
> 
> Probably the hairiest part of Jolt was managing all of the shapes. Conceptually, LÖVR models physics bodies as having a list of shapes attached to them. The shapes can each have an offset, and the body can have a custom mass/COM. Whenever someone adds/removes a shape, or adjusts the mass stuff, we have to do a bunch of gymnastics: promote to a `MutableCompoundShape` if needed, switch to an `EmptyShape` if last shape was removed, wrap in an `OffsetCenterOfMassShape` to preserve their custom center of mass if needed, notify all the joints if COM changed, etc. etc. There's a bit of a combinatorial explosion of cases that need to be handled to keep things consistent, and I'm still not sure that I got it all right, see [here](https://github.com/bjornbytes/lovr/blob/0ab47ba8f13a06419588d6a4c295ab58f20aea59/src/modules/physics/physics.c#L1029-L1060) for an example in the `lovrColliderAddShape` function. Part of me wishes that Jolt modeled bodies as having a list of shapes, and that the compound/decorator shapes weren't necessary, but it would be less flexible, make structs bigger, and I imagine this design is necessary due to thread safety.
> 
> Also, having a small library size was really important. I did some measurements and found that the object stream was inflating code size a lot, but then you added `ENABLE_OBJECT_STREAM` a few days later 😄. Currently, LÖVR also doesn't use characters, vehicles, ragdolls, or soft body, and it would be nice to have the option of disabling those to save some size. I wasn't able to figure out how to yank them out cleanly though, and it didn't seem like the savings were going to be as impressive as object stream, so I moved on.
> 
> But overall, Jolt has been really great!

It's been a joy and honor follow your progress and merging your PR, feel free to talk to me if you need something. 

Cheers, 
Amer

jrouwe:
Thanks for the feedback!

> Part of me wishes that Jolt modeled bodies as having a list of shapes, and that the compound/decorator shapes weren't necessary, but it would be less flexible, make structs bigger, and I imagine this design is necessary due to thread safety.

Yes, most shapes are non-mutable in order to ensure that they're thread safe. MutableCompoundShape is not thread safe if you randomly add/remove shapes to it (we always create a new MutableCompoundShape to ensure things are safe). 

They're made out of small building blocks because we were really low on memory on PlayStation 4 and we saved about 30% of memory when switching to Jolt because of it. I've seen multiple people struggling with this (see e.g. https://github.com/jrouwe/JoltPhysics/discussions/1165), so I'll put it on my list to come up with maybe a helper construct to ease creation.

> I did some measurements and found that the object stream was inflating code size a lot, but then you added ENABLE_OBJECT_STREAM a few days later 😄

I happened to be looking at the size of the JavaScript library at the exact same time 😄.

> Currently, LÖVR also doesn't use characters, vehicles, ragdolls, or soft body, and it would be nice to have the option of disabling those to save some size. I wasn't able to figure out how to yank them out cleanly though, and it didn't seem like the savings were going to be as impressive as object stream, so I moved on.

My conclusions were the same after removing the object stream stuff.

* Ragdolls/Vehicles: The only thing that ties them into the library should be the `RegisterTypes` function. So if you would copy paste that function and remove the classes that you don't need, I would expect dead code elimination to do the rest.
* Character: There are no references from the library onto the Character classes, so if you're not using it I would expect dead code elimination to remove it.
* Softbody: This one is very tricky to remove as it is intertwined into the job graph. I would suggest integrating cloth into LÖVR to make sure you make the most of these bytes 😄.

In any case, if you need any other features let me know!

PR #1540, opened 2025-03-01
Improved stability of cylinder contacts

jrouwe:
* When the inDirection is more than 5 degrees from vertical, align the vertices so that 1 of the vertices points towards inDirection in the XZ plane. This ensures that we always have a vertex towards max penetration depth.
* Removed duplicate if in ManifoldBetweenTwoFaces


sonukaloshiya24:
feature/cyclinder_stability

PR #1592, opened 2025-04-08
Set JPH_DVECTOR_ALIGNMENT to 32 for LoongArch

stdmnpkg:
original value only works for gcc with single precision, this change fix clang build and double precision build.

tested on my LoongArch machine with gcc 14, clang 18 and clang 20

### DIFF
--- a/Jolt/Core/Core.h
+++ b/Jolt/Core/Core.h
@@ -229,7 +229,7 @@
 		#define JPH_CPU_ADDRESS_BITS 32
 	#endif
 	#define JPH_VECTOR_ALIGNMENT 16
-	#define JPH_DVECTOR_ALIGNMENT 8
+	#define JPH_DVECTOR_ALIGNMENT 32
 #elif defined(__e2k__)
 	// E2K CPU architecture (MCST Elbrus 2000)
 	#define JPH_CPU_E2K


stdmnpkg:
it's worth noting some test failed with clang build. see https://github.com/loongson-community/discussions/issues/89

jrouwe:
I made some modifications to your change. Basically DVec3 should be 8 byte aligned on loongarch since there is currently no SIMD support there. The static_asserts were a bit trigger happy so I toned them down.

stdmnpkg:
this does not compile with clang 18

```
error: requested alignment is less than minimum alignment of 16 for type 'JPH::DMat44'
   12 | class [[nodiscard]] alignas(JPH_DVECTOR_ALIGNMENT) DMat44
      |                     ^
1 error generated.
```

jrouwe:
Ah, ok I see why (it's difficult to test this when I'm only able to compile with gcc). Let me fix it.

jrouwe:
Does this work?

stdmnpkg:
clang single precision build success, clang double precision build failed
```
Jolt/Physics/Body/Body.h:36:32: error: requested alignment is less than minimum alignment of 16 for type 'JPH::Body'
   36 |         JPH_EXPORT_GCC_BUG_WORKAROUND alignas(JPH_RVECTOR_ALIGNMENT)
      |                                       ^
1 error generated.
```

jrouwe:
That one should be fixed too...

stdmnpkg:
yes, this one works

jrouwe:
Thanks for reporting and helping to get this fixed!

PR #1599, opened 2025-04-10
Adds CMake option to use External Profiler

jeslaspravin:
Ability to use the external profiler using CMake config. 
Use an external profiler set using ProfileStartMeasurement and ProfileEndMeasurement to profile. 
The option is available only when profiling is enabled using PROFILER_IN_DEBUG_AND_RELEASE or PROFILER_IN_DISTRIBUTION.

Exposes the preprocessor `JPH_EXTERNAL_PROFILE` as CMake option. Does not make any functionality changes.

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -92,10 +92,9 @@ option(PROFILER_IN_DEBUG_AND_RELEASE "Enable the profiler in Debug and Release b
 # Note that enabling this reduces the performance of the library.
 option(PROFILER_IN_DISTRIBUTION "Enable the profiler in all builds" OFF)
 
-# Ability to use the external profiler using CMake config.
 # Use external profiler set using ProfileStartMeasurement and ProfileEndMeasurement to profile.
 # Option is available only when profiling is enabled using PROFILER_IN_DEBUG_AND_RELEASE or PROFILER_IN_DISTRIBUTION.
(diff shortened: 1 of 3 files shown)


jrouwe:
Could you rename the option from `USE_EXTERNAL_PROFILER` to `JPH_USE_EXTERNAL_PROFILER`? Eventually I will prefix all cmake options to avoid collisions with other projects, but I'll probably do this on a major version bump.

jeslaspravin:
Done

jrouwe:
Hmmm, I see some further inconsistencies: `JPH_PROFILE_ENABLED` and `JPH_EXTERNAL_PROFILE` are supposed to be mutually exclusive. So I think the updated text in `README.md` doesn't make sense and I think `Jolt.cmake` also needs to be updated to make the settings mutually exclusive.

jeslaspravin:
Right, I will make change to align with mutual exclusivity priority implemented at  https://github.com/jrouwe/JoltPhysics/blob/master/Jolt/Core/Profiler.h#L16. Will that be alright?

jeslaspravin:
Fixed, now only one of `JPH_EXTERNAL_PROFILE`, `JPH_PROFILE_ENABLED` gets defined. Selecting one of `JPH_EXTERNAL_PROFILE` or `JPH_PROFILE_ENABLED` preprocessor define based on `JPH_USE_EXTERNAL_PROFILE`

jrouwe:
Thanks!

Issue #1600, opened 2025-04-11
clang++-20: error: overriding '-ffp-model=precise' option with '-ffp-contract=off' [-Werror,-Woverriding-option]


tksuoran:
There already is this:
```cmake
# emcc complains with: warning: overriding '-ffp-model=precise' option with '-ffp-contract=off' [-Woverriding-option],
# but this is exactly what we want.
if (EMSCRIPTEN)
	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-overriding-option")
endif()
```
I guess that could be extended to cover clang 20 as well?

jrouwe:
Can you check if this fixes it?

https://github.com/jrouwe/JoltPhysics/commit/3ce95bb8455f11bdd30230ec2334d8dcd63adf24

tksuoran:
Sorry, I forgot to nention, the pull request change does fix the issue for me.

jrouwe:
Closing as the fix was already applied on master

PR #1601, opened 2025-04-12
[clang] Handle ffp-model=precise with ffp-contract=off

tksuoran:
Specifying -ffp-contract=off after -ffp-model=precise is intentional. clang emits diagnostic warning that the override has occurred. This should not be treated as error.

Fix for #1600


jrouwe:
I double checked: You made your change on an older version of Jolt, the latest version already contains a change that is almost the same as yours:

https://github.com/jrouwe/JoltPhysics/blob/829d246742e8b83635b0927414f03d6b910b87f1/Build/CMakeLists.txt#L248-L250

tksuoran:
Ah, apologies, my bad. Yes, can close. Thanks!

PR #1604, opened 2025-04-13
Add parameter to shrink rest length of edge constraints

elzewyr:
Scaling down the rest lengths of edge constraints produces a shrinking effect that is useful for creating constantly tightened cloth, like sails or bat wings. This PR adds `mEdgeRestLengthMultiplier` parameter that can be set at the creation of the softbody and implemented as an adjustable parameter in Godot or other projects that uses Jolt, similar to Shrinking Factor in Blender's cloth simulation modifier.

ShapesTest with `mEdgeRestLengthMultiplier = 0.5`:
image


jrouwe:
Isn't it better to premultiply all edge lengths after creating the `SoftBodySharedSettings` object?

I.e.:

```C++
shared_settings = new SoftBodySharedSettings;
...fill in the settings...
shared_settings->CalculateEdgeLengths();
for (SoftBodySharedSettings::Edge &e : shared_settings->mEdges)
    e.mRestLength *= factor;
```

Now you pay the cost of an extra multiplication for every vertex for every solver iteration for every simulation step.

elzewyr:
Yes, this appears to be a superior approach, I will implement it in Godot directly. Thank you for pointing this out.

Issue #1633, opened 2025-05-10
Build Error: Type Conversion Warnings Treated as Errors in Version 5.3.0


STIXTheMachine:
When attempting to build the samples in version 5.3.0, the compilation fails due to warnings being treated as errors. Specifically, the issues are related to type conversion warnings from size_t to JPH::uint and from size_t to uint32_t.

Multiple files in the TestFramework are affected:
`TestFramework/UI/UIManager.cpp` (lines 47-48)
`TestFramework/Renderer/Font.cpp` (line 153)
`TestFramework/Renderer/VK/PipelineStateVK.cpp` (lines 145, 150)
`TestFramework/Renderer/DebugRendererImp.cpp` (lines 32, 49-51, 56-58)
`TestFramework/Renderer/VK/RendererVK.cpp` (lines 400, 420, 432, 553, 601, 740, 951)

Build environment:
- Attempted using Visual Studio 2022 Community and CLion 2025.1
- CMake 3.26.3
- Windows 10 Pro
- Attempted using all build configurations

A complete error log is available in [this Gist](https://gist.github.com/STIXTheMachine/b8f9b5af90c7f190b264083d4c1c5da8)

This can probably just be fixed by just suppressing these errors around the affected lines since it's not actual library code.

Has anyone else experienced this? I'd be happy to patch this real quick if just suppressing the warnings is sufficient.

jrouwe:
Looks like you're using an older verson of MSVC, the behavior of `std::size` changed on MSVC 17.10 (according to your gist you're on a lower version). 

See godbolt:
[...]

PR #1634, opened 2025-05-10
Add CMake flag for Unreal Engine compatibility

STIXTheMachine:
There has been growing interest in integrating Jolt physics with Unreal Engine. One small issue with this is that when building with MSVC, Jolt defines `_HAS_EXCEPTIONS = 0`  to disable exceptions in the STL for performance reasons. However this conflicts with Unreal Engine's build settings, causing compilation to fail. A full log of the errors this causes can be found in [this Gist.](https://gist.github.com/STIXTheMachine/126021ecf5e53da0b441a89c01186654)

This issue has been resistant to every solution I've tried from within Unreal Engine's build configuration options.

This PR introduces the CMake option ~~`WITH_UNREAL_ENGINE`~~ `JPH_WITH_UNREAL_ENGINE` which disables the `_HAS_EXCEPTIONS` redefinition when enabled. This option defaults to OFF, so existing builds are unaffected.

~~Unit tests pass successfully, note however that samples do not build build due to issue #1633 .~~
Unit tests and samples are confirmed to compile and pass.

### DIFF
--- a/Build/CMakeLists.txt
+++ b/Build/CMakeLists.txt
@@ -111,7 +111,7 @@ option(ENABLE_INSTALL "Generate installation target"  ON)
 
 # Enable Unreal Engine integration
 # Primarily stops us from redefining _HAS_EXCEPTIONS which causes conflicts when building with Unreal
-option(WITH_UNREAL_ENGINE "Enable Unreal Engine integration" OFF)
+option(JPH_WITH_UNREAL_ENGINE "Enable Unreal Engine integration" OFF)
 
 include(CMakeDependentOption)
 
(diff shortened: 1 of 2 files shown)


STIXTheMachine:
In reviewing the PR for #1599 I see that there is a preference for CMake options to have the `JPH_` prefix going forward. I have renamed the flag from `WITH_UNREAL_ENGINE` to `JPH_WITH_UNREAL_ENGINE`.

jrouwe:
Hello,

I'm not a big fan of introducing settings specific for projects that use Jolt. I've created an alternative solution:

https://github.com/jrouwe/JoltPhysics/pull/1635/files

Will this work for you?

STIXTheMachine:
I can confirm that #1635 solves this issue. Thank you!

PR #1635, opened 2025-05-10
Moved the definition of _HAS_EXCEPTIONS from cmake to Core.h

jrouwe:


STIXTheMachine:
I can confirm that this solves the issue described in #1634 , I have now closed that issue.

PR #1642, opened 2025-05-16
fix missing malloc.h include in clang-cl mode (clang in msvc mode)

virt00l:
The issue manifest itself in windows-on-arm build (`--target=aarch64-pc-windows-msvc`)

Related
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170


jrouwe:
Thanks!

PR #1645, opened 2025-05-20
README.md: update weblink to the Jolt-JNI project (Java/Kotlin bindings)

stephengold:
I created a documentation website for the Jolt-JNI project,
and I'd like for the new site to become the main entry point into the project.
So I'm requesting a change to the weblink in your project's README.

I'm also adding mention of Kotlin. Jolt JNI works equally well with Java and Kotlin, so it's a nice freebie.
Presumably Jolt JNI will also work with Scala, Groovy, and Clojure.

I've made good progress on Jolt JNI since September, when the weblink was created.
Vehicles and soft bodies work great, and there's an OpenGL-based visualization add-on that works
equally well on Windows, Linux, and macOS.

I'm happy to answer any questions you have about the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ If you're interested in how Jolt scales with multiple CPUs and compares to other
 
 * C [here](https://github.com/amerkoleci/joltc), [here](https://github.com/zig-gamedev/zphysics/tree/main/libs/JoltC) and [here](https://github.com/SecondHalfGames/JoltC/)
 * [C#](https://github.com/amerkoleci/JoltPhysicsSharp)
-* [Java](https://github.com/stephengold/jolt-jni)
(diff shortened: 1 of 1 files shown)


jrouwe:
You put a lot of effort in that documentation! (looks like it could be more than what I have done for Jolt itself)

If I may make a suggestion: It would be cool to add some movies of the examples you have created on the main page as a teaser.

stephengold:
Thanks for browsing the documentation site. It's based on a similar site that I created for another physics library. It took a lot of effort, but if Jolt JNI proves popular it may save time in the long run.

I agree that some video, or even just more graphics, would make the site more appealing. I appreciate your suggestion. I'll add it to my to-do list.

PR #1680, opened 2025-06-23
Potential Bug in Contact Constraint Manager

nicholaszollinger:
The penetration depth calculation in `ContactConstraintsManager::GetContactsFromCache()` was using the first relative contact points for each iteration of the loop.


nicholaszollinger:
Awesome, glad to help!

PR #1692, opened 2025-07-07
Tumble-Lock Pattern

JKurzer:
Hey JRouwe!

This is a very very small change but can significantly improve performance in certain high contention situations. Because many of our locks are very very short-lived, this "tumble lock" pattern where you spin 2-3 times before actually locking can save milliseconds of lost CPU time as well as prevent some nasty cache thrash.

I suspect there's a more beautiful way to implement it, that's more idiomatic with jolt's style. Let me know if you have a way you'd _prefer_ to see this done, and I'll revise to fit. Sorry to propose this against master, I'll swap it to your preferred branch once it's revised to spec.

--J

### DIFF
--- a/Jolt/Physics/PhysicsLock.h
+++ b/Jolt/Physics/PhysicsLock.h
@@ -54,7 +54,13 @@ public:
 	static inline void			sLock(LockType &inMutex JPH_IF_ENABLE_ASSERTS(, PhysicsLockContext inContext, EPhysicsLockTypes inType))
 	{
 		JPH_IF_ENABLE_ASSERTS(sCheckLock(inContext, inType);)
-		inMutex.lock();
+		if (!inMutex.try_lock()) //sleeping against a lock is the most expensive possible operation.
+		{
+			if (!inMutex.try_lock()) //so we spin twice first.
+			{
+				inMutex.lock(); //that might sound random, but actually, most locks are supremely ephemeral.
+			}
+		}
 	}
 
 	template <class LockType>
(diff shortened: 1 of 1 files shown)


JKurzer:
It is primarily an issue at extremely high object counts. Profile and
assert are disabled, but I will double check. I'll also gather some
metrics.

The locks should be invisible most of the time. I'm using body counts in
the 10k+ range.

That said, this is strictly better than a pure lock, on almost all
platforms. I don't have a full set of dev kits at the moment, so if you
wanted to run a bench on PS5, that might be informative.

jrouwe:
We usually have 100K+ bodies in the scene, so my guess is that it's something particular you're doing. In general, `PhysicsSystem::Update` locks all mutexes at the start and unlocks them at the end so within the update there should be no contention. The only contention I can think of can come from doing collision checks while inserting/removing bodies from different threads. Or different threads inserting/removing bodies at the same time.

JKurzer:
That's genuinely worrisome. I'll take a pass over the examples and see if
I've done something boneheaded. It seems likely.

nanokatze:
Most platforms' mutexes already do this kind of optimization internally

jrouwe:
Before including this I would prefer to see for myself that this actually improves performance, if possible in the Samples app (in an existing scene or a new one). Also I would still like to know what the callstacks are of the things that are producing contention for you.

JKurzer:
> ?

Been a bit busy and had a couple family things come up, but we should be circling back and providing either test evidence or closing this PR soon-ISH

PR #1770, opened 2025-09-28
Export IslandBuilder class by adding JPH_EXPORT

Megafunk:
This is so that I can create new island-affecting constraints similar to what VehicleConstraint::BuildIslands does when using Jolt as a DLL!

There might be more non-exported classes but I haven't encountered any besides this so far in my project. Big thanks to JRouwe and other contributors for their time and effort put into this repo and answering questions/PRs.

### DIFF
--- a/Jolt/Physics/IslandBuilder.h
+++ b/Jolt/Physics/IslandBuilder.h
@@ -15,7 +15,7 @@ class TempAllocator;
 //#define JPH_VALIDATE_ISLAND_BUILDER
 
 /// Keeps track of connected bodies and builds islands for multithreaded velocity/position update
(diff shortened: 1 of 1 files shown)


Megafunk:
I then tried to make my own constraint that affects island building similar to how the vehicle constraint does by linking bodies it is nearby or has contacts with. The idea being that I could sort of blob the characters together and then let Jolt handle moving them.
I suppose making a brand new constraint is definitely not something users will be doing all of the time so if I just have to add my new one in the same module as Jolt that's fine. I personally think that being able to add new constraints to a physics library is useful but perhaps something relies on the set of constraints being unchanged that I am unaware of.

PR #1809, opened 2025-11-16
Make assert in BodyManager::ActivateBodies only for dynamic bodies

IvanZinkevich:
When you call ActivateBodies with static bodies as well, the assert is fired. I put it under if condition.


jrouwe:
This assert is there for a good reason. You can't call ActivateBodies on something that is not in the broad phase. If you go through this function anyway then it will crash and burn somewhere later.

PR #1827, opened 2025-11-25
improve temp alloc a bit (non aligned allocs)

virt00l:
If JPH_RVECTOR_ALIGNMENT is small enough we could use simpler/less overhead Allocate/Free code path instead of aligned one.

Basically same logic as in
https://github.com/jrouwe/JoltPhysics/commit/ff05612c89b1ba0a9c9fca7bcc3bacd34b08dfc9#diff-88beae258ce8db547ebd0dfb2bad98ad6c995b211dd368316a4a816d4bffd796R47


jrouwe:
Thanks!

PR #1847, opened 2025-12-28
Added interface to run compute shaders on the GPU with implementations for DX12, Vulkan and Metal.

jrouwe:
Currently it is only used by a unit test, further users of this system will follow later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
       run: ctest --output-on-failure --verbose
 
   linux_clang_tsan:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
     name: Linux Clang Sanitizers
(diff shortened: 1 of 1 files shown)


Jolt/Core/IncludeWindows.h
```diff
@@ -0,0 +1,35 @@
...
+	#define WIN32_LEAN_AND_MEAN
+#endif
+
+#ifndef JPH_COMPILER_MINGW
+	#include <Windows.h>
+#else
+	#include <windows.h>
+#endif
+
+#undef min // We'd like to use std::min and max instead of the ones defined in windows.h
```
SirLynix: Why not `#define NOMINMAX` instead?

jrouwe: Because I didn't know that that existed :). I'll change it.

PR #1957, opened 2026-03-22
Created SSE4.1 version of Vec3::GetNormalizedPerpendicular

jrouwe:
This function is used by ContactConstraintBase::GetTangents and improves performance by 1.5% in the Pyramid performance test

### DIFF
--- a/Jolt/Math/Vec3.inl
+++ b/Jolt/Math/Vec3.inl
@@ -1216,6 +1216,35 @@ float Vec3::ReduceMax() const
 
 Vec3 Vec3::GetNormalizedPerpendicular() const
 {
+#ifdef JPH_USE_SSE4_1
+	// Compute -value and |value| and check if |x| > |y|
+	__m128 neg_val = _mm_sub_ps(_mm_setzero_ps(), mValue);
+	__m128 abs_val = _mm_max_ps(mValue, neg_val);
+	__m128 abs_x = _mm_shuffle_ps(abs_val, abs_val, _MM_SHUFFLE(0, 0, 0, 0));
+	__m128 abs_y = _mm_shuffle_ps(abs_val, abs_val, _MM_SHUFFLE(1, 1, 1, 1));
+	__m128 x_gt_y = _mm_cmpgt_ps(abs_x, abs_y);
+
+	// |x| > |y|: perpendicular is [z, 0, -x, 0]
+	__m128 case_x_gt_y = _mm_shuffle_ps(mValue, neg_val, _MM_SHUFFLE(0, 0, 2, 2));
+	case_x_gt_y = _mm_and_ps(case_x_gt_y, _mm_castsi128_ps(_mm_set_epi32(0, -1, 0, -1)));
+
+	// |x| <= |y|: perpendicular is [0, z, -y, 0]
+	__m128 case_x_le_y = _mm_shuffle_ps(mValue, neg_val, _MM_SHUFFLE(1, 1, 2, 2));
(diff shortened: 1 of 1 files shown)


Tomarty:
I ported this function to C a few months back for my own math library and came up with a simpler formulation of the algorithm for SSE2 and the fallback. Bit-exact against the inputs I fuzzed, though can diverge when the squares hit infinity. Sharing in case it's useful.

```
Vec3 Vec3::GetNormalizedPerpendicular() const
{
#if defined(JPH_USE_SSE)
    // Build both perpendicular candidates without explicit masking:
    //   perp_x = [z, 0, -x, 0]  (used when |x| > |y|)
    //   perp_y = [0, z, -y, 0]  (used when |x| <= |y|)
    __m128 zero = _mm_setzero_ps();
    __m128 neg = _mm_sub_ps(zero, mValue); // alt: xor with sign mask
    __m128 perp_x = _mm_shuffle_ps(_mm_unpackhi_ps(mValue, zero), neg, _MM_SHUFFLE(0, 0, 1, 0));
    __m128 perp_y = _mm_shuffle_ps(_mm_unpackhi_ps(zero, mValue), neg, _MM_SHUFFLE(1, 1, 1, 0));

    // Compare squared components instead of absolute values (saves the abs computation).
    __m128 sq = _mm_mul_ps(mValue, mValue);
    __m128 xx = _mm_shuffle_ps(sq, sq, _MM_SHUFFLE(0, 0, 0, 0));
    __m128 yy = _mm_shuffle_ps(sq, sq, _MM_SHUFFLE(1, 1, 1, 1));
    __m128 zz = _mm_shuffle_ps(sq, sq, _MM_SHUFFLE(2, 2, 2, 2));
    __m128 x_gt_y = _mm_cmpgt_ps(xx, yy);

    // Select perpendicular based on |x| > |y|.
#if defined(JPH_USE_SSE4_1) && !defined(JPH_PLATFORM_WASM) // _mm_blendv_ps has problems on FireFox
    __m128 result = _mm_blendv_ps(perp_y, perp_x, x_gt_y);
#else
    __m128 result = _mm_or_ps(_mm_and_ps(x_gt_y, perp_x), _mm_andnot_ps(x_gt_y, perp_y));
#endif

    // Normalize. Since the result has only two nonzero components; one of x²/y² plus z²; the squared length is max(xx, yy) + zz. All lanes of the sqrt input are identical.
    __m128 len = _mm_sqrt_ps(_mm_add_ps(_mm_max_ps(xx, yy), zz));
    return sFixW(_mm_div_ps(result, len));
#else
    float x = mF32[0], y = mF32[1], z = mF32[2];
    float xx = x * x, yy = y * y, zz = z * z;
    // e.g. `0.0f - x` does not produce signed zero and is consistent with `_mm_sub_ps(zero, mValue)`
    // alt: 0.0f - x, 0.0f - y
    Vec3 b(z, 0.0f, -x);
    Vec3 c(0.0f, z, -y);
    return (xx > yy ? b : c) / sqrt(zz + (xx > yy ? xx : yy));
#endif
}
```

P.S. Flagged both negation styles with alt: comments. They can differ in the sign bit of zero components. minps/maxps already scramble zero signs relative to NEON, so I figured it's fine.

jrouwe:
I can confirm that this version is approx. one third faster. Thanks for that!

I've created a PR for this #1979 

> They can differ in the sign bit of zero components. minps/maxps already scramble zero signs relative to NEON, so I figured it's fine.

What do you mean with this? Do you mean that the old version already had a problem with -0.0? (looking at it, I think it did, which is bad). I think I fixed it though.

PR #1973, opened 2026-04-11
Load Vulkan dynamically instead of linking it as a static library

jrouwe:


stephengold:
Is Vulkan compute supported on Android platforms? The changes this PR made to ComputeSystemVKImpl.cpp suggest it's not supported.

jrouwe:
I haven't tested it on Android. Perhaps it can just use the Linux code path? (If you have a way of verifying this then I'd be happy to take the change)

stephengold:
I'll attempt verification.

PR #1977, opened 2026-04-17
Added acceleration-mode velocity drive to SixDOFConstraint motors

khiner:
A velocity motor whose MotorSettings::mSpringSettings has damping > 0 and no stiffness/frequency now produces a soft, mass-normalized velocity drive (a = -damping * v_err) via an implicit-Euler soft constraint. Previously the damping was ignored and the motor acted as a hard velocity constraint. Motors with damping == 0, and position motors, are unchanged.

Added unit test TestSixDOFMotorVelocityAcceleration covering both translation and rotation motor paths, both spring modes, and two body sizes to verify mass/inertia independence.

Disclosure: I developed this with the help of Claude Code.

### DIFF
--- a/Docs/APIChanges.md
+++ b/Docs/APIChanges.md
@@ -6,6 +6,7 @@ Changes that make some state saved through SaveBinaryState from a prior version 
 
 ## Changes between v5.5.0 and latest
 
+* 20260417 - `SixDOFConstraint` velocity motors with `MotorSettings::mSpringSettings` damping > 0 and no stiffness/frequency now act as soft acceleration-mode velocity drives (`a = -damping * v_err`) rather than hard velocity constraints (where the damping was previously ignored). Motors with damping == 0 are unchanged.
 * 20260410 - Fixed contact callbacks for body with motion quality LinearCast vs a soft body. Previously, the contacts would be reported accidentally through the regular ContactListener. Now they're properly reported through the SoftBodyContactListener. (63765d19bae439ea4a9f93d186d6f1d94029229b)
 * 20260307 - *SBS* - Added support for HeightFieldShapeSettings::mBitsPerSample > 8 which adds 1 byte to the binary serialization format and renders it incompatible with previous saved data. (449b645b71a7a47aa0d7bdcb5f9c197f1ddff5b0)
 * 20253012 - Added interface to run compute shaders on the GPU with implementations for DX12, Vulkan and Metal. These interfaces can be disabled by setting JPH_USE_DX12, JPH_USE_VK and JPH_USE_MTL to OFF. To build on macOS, you'll need to have dxc and spirv-cross installed. The easiest way to install them is by installing the Vulkan SDK. (5ac132df689fbf88da618181b0f1f73fca8bb1b4)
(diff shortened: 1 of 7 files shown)


jrouwe:
Thanks! It will take me some time to validate this change.

In the mean time: What is your use case for this and can't you use the built in friction of the constraints (e.g. `SixDOFConstraintSettings::mMaxFriction`)? Obviously that will not give you the exact same formula as what you've implemented here.

khiner:
Thanks for taking a look!

My use case is the [KHR_physics_rigid_bodies Robot_skinned](https://github.com/eoineoineoin/glTF_Physics/blob/master/samples/Robot_skinned/Robot_skinned.gltf) glTF sample. It defines three angular drives in `mode: "acceleration"`: two of them (legs and cog) have `stiffness: 0` with `damping > 0` and a non-zero `velocityTarget`. The [spec's `joint.drive` section](https://github.com/eoineoineoin/glTF_Physics/tree/master/extensions/2.0/Khronos/KHR_physics_rigid_bodies#joints) defines the drive as:
  > proportional to `stiffness * (positionTarget - positionCurrent) + damping * (velocityTarget - velocityCurrent)` … in `acceleration` mode … scaled by the effective mass of the driven degree of freedom.

So that case reduces to `a = damping * (velocityTarget - velocityCurrent)`, which I couldn't express with Jolt as-is.

* Video of the robot with this patch:
https://github.com/user-attachments/assets/8318deac-e960-40f5-809b-27aeb8434dfb

* Video with `mMaxFriction`:
https://github.com/user-attachments/assets/d73a3443-4c7f-4fe7-8e36-4f7e2caf04d5

Since it's a passive friction cap, the leg and cog drives don't apply force toward their `velocityTarget` and the robot doesn't move.

However, I went back and realized for this scene, on master, letting Jolt's velocity motor handle these drives
looks visually just about identical (with `EMotorState::Velocity` ignoring `mSpringSettings` entirely and a hard velocity constraint at `velocityTarget`). I think because the damping values are large enough, even though it ignores the `damping` value it seems to look fine in this case:
https://github.com/user-attachments/assets/5f723aa1-cdd8-4b49-8c21-a363d1b1ae28

Still, dropping the damping prevents implementing KHR_physics_rigid_bodies exactly. Instantaneous lock to `velocityTarget` instead of a ramp would show better when damping is small or the driven body is light.

Anyway, not as blocked as I thought on this one, so feel free to do what you'd like with this PR, thanks for looking :)

jrouwe:
I've looked at your change a little closer and created an alternative version of it here: #1990 (I marked you co-author).

The main problems your implementation had were:
* Broke compatibility when frequency was not specified but damping was (I created `EMotorState::PositionAndVelocity` for this - this mimics how the `KHR_physics_rigid_bodies` spec treats springs)
* The damping parameter switched to acceleration mode only when frequency is zero (I created `ESpringMode::MassNormalizedStiffnessAndDamping` for this)
* It was only implemented on the `SixDOFConstraint` (now all constraints that have motors support it)

I think with these new modes you should be able to implement almost the entire `KHR_physics_rigid_bodies` spec. The only thing that I think is still missing is the angular spring limits for `SixDOFConstraint`.

This change needs more testing and I need to double check that I didn't change existing behavior for the case where stiffness = 0 and damping > 0.

khiner:
Thanks so much for taking the time to turn this into a proper implementation! I updated to use it and it seems to work great, and improves the motor behavior visually. Closing this.

PR #1983, opened 2026-04-23
ComputeSystemVKImpl.cpp:  add support for Android platforms

stephengold:
This PR adds support for Vulkan compute servers on Android platforms.
I tested this on an emulator, and it worked as expected.


jrouwe:
Thanks!

PR #1990, opened 2026-04-27
Added support for glTF KHR_physics_rigid_bodies constraint motors

jrouwe:
* Added constraint motors that can drive both position & velocity
* Allow spring to only have damping and no stiffness


jrouwe:
@khiner Note that this PR is merged now and the branch deleted. I'm not sure if the dependency you created in https://github.com/khiner/MeshEditor/commit/da7d303c0ae0df8db9dd04a639148f7ffc86b80d still exists now, you may want to update to target `master` again.

PR #1991, opened 2026-04-28
Use vfsqrt in place of sqrt libm in RVV code.

topperc:
The sqrt libm function requires errno to be set if the input is negative. For performance, clang and gcc partially inline the sqrt libcall by emitting a scalar sqrt instruction and a compre+branch to check for the error condition. If the error occurs, the sqrt function in libm will be called to set errno. For RVV, this call to sqrt requires the vector registers to be spilled and reloaded.

I suspect in most cases this error condition doesn't occur so this extra code just bloats the binary. The extra code can be disabled by passing -fno-math-errno, but I don't know if that's desirable for this library.

This patch instead uses the vfsqrt instrinsic in place of the library call. I've changed the interface to RVVSumElementsFloat32x4 to return the result in element 0 of a vector. The callers will move to a scalar register if necessary. If the scalar result needs to be splatted, I've used vrgather.vi.


jrouwe:
Thanks! I actually have a change lying around that replaces all usages of `std::sqrt` with my own `Sqrt` for the same reason. I'll see if I can dust that off.

jrouwe:
See #1992

PR #1992, opened 2026-04-28
Replace std::sqrt with Sqrt

jrouwe:
std::sqrt is required to set errno in case of an error, but we don't care about that. This sometimes causes the function to not be inlined.

### DIFF
--- a/Jolt/Math/Math.h
+++ b/Jolt/Math/Math.h
@@ -57,6 +57,34 @@ JPH_INLINE constexpr T Square(T inV)
 	return inV * inV;
 }
 
+/// Take the square root of a float value
+JPH_INLINE float Sqrt(float inV)
+{
+#ifdef JPH_USE_SSE
+	return _mm_cvtss_f32(_mm_sqrt_ss(_mm_set_ss(inV)));
+#elif defined(JPH_USE_NEON)
+	return vget_lane_f32(vsqrt_f32(vdup_n_f32(inV)), 0);
+#elif defined(JPH_USE_RVV)
+	return __riscv_vfmv_f_s_f32m1_f32(__riscv_vfsqrt_v_f32m1(__riscv_vfmv_v_f_f32m1(inV, 1), 1));
+#else
+	return std::sqrt(inV);
+#endif
+}
+
+/// Take the square root of a double value
+JPH_INLINE double Sqrt(double inV)
+{
+#ifdef JPH_USE_SSE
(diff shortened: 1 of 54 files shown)


Jolt/Math/Math.h
```diff
@@ -57,6 +57,34 @@ JPH_INLINE constexpr T Square(T inV)
...
+
+/// Take the square root of a double value
+JPH_INLINE double Sqrt(double inV)
+{
+#ifdef JPH_USE_SSE
+	return _mm_cvtsd_f64(_mm_sqrt_sd(_mm_undefined_pd(), _mm_set_sd(inV)));
+#elif defined(JPH_USE_NEON)
+	return vget_lane_f64(vsqrt_f64(vdup_n_f64(inV)), 0);
+#elif defined(JPH_USE_RVV)
+	return __riscv_vfmv_f_s_f64m1_f64(__riscv_vfsqrt_v_f64m1(__riscv_vfmv_v_f_f64m1(inV, 1), 1));
```
topperc: It's probably better to use the scalar sqrt instruction here, but we don't have an intrinsic for that today. Is it possible to use inline assembly here?

jrouwe: Yes, as long as it compiles on gcc and clang. I don't have time to look up the syntax tonight, but if you know what it should be please let me know.

Jolt/Math/Math.h
```diff
@@ -57,6 +57,34 @@ JPH_INLINE constexpr T Square(T inV)
...
 
+/// Take the square root of a float value
+JPH_INLINE float Sqrt(float inV)
+{
+#ifdef JPH_USE_SSE
+	return _mm_cvtss_f32(_mm_sqrt_ss(_mm_set_ss(inV)));
+#elif defined(JPH_USE_NEON)
+	return vget_lane_f32(vsqrt_f32(vdup_n_f32(inV)), 0);
+#elif defined(JPH_USE_RVV)
+	return __riscv_vfmv_f_s_f32m1_f32(__riscv_vfsqrt_v_f32m1(__riscv_vfmv_v_f_f32m1(inV, 1), 1));
```
topperc:
```
  float res;
  asm("fsqrt.s %0, %1" : "=f"(res) : "f"(inV));
  return res;
```

Jolt/Math/Math.h
topperc:
```
  double res;
  asm("fsqrt.d %0, %1" : "=f"(res) : "f"(inV));
  return res;
```

PR #2010, opened 2026-05-08
Added ReduceAdd() to consolidate horizontal-add pattern

asenkew:
Hello, Jorrit,

here is optimization of dot-product giving performance uplift.

### DIFF
--- a/Jolt/Math/Vec3.inl
+++ b/Jolt/Math/Vec3.inl
@@ -954,12 +954,7 @@ float Vec3::LengthSq() const
 
 float Vec3::Length() const
 {
-#if defined(JPH_USE_NEON)
-	// Minor differences in NaN handling make sqrt() produce
-	// a few extra instructions. Hence, we need to hand-roll
-	// this one for efficiency.
-	return vget_lane_f32(vsqrt_f32(vdup_n_f32(LengthSq())), 0);
-#elif defined(JPH_USE_RVV)
+#if defined(JPH_USE_RVV)
 	const vfloat32m1_t zeros = __riscv_vfmv_v_f_f32m1(0.0f, 3);
 	const vfloat32m1_t v = __riscv_vle32_v_f32m1(mF32, 3);
 	const vfloat32m1_t mul = __riscv_vfmul_vv_f32m1(v, v, 3);
@@ -1014,9 +1009,6 @@ Vec3 Vec3::NormalizedOr(Vec3Arg inZeroValue) const
 	Type shuf = _mm_movehdup_ps(mul);
 	Type sums = _mm_add_ps(mul, shuf);
 	shuf = _mm_movehl_ps(mul, mul);
-#ifdef JPH_CROSS_PLATFORM_DETERMINISTIC
-	shuf = _mm_add_ps(_mm_setzero_ps(), shuf);
-#endif
 	sums = _mm_add_ps(sums, shuf);
 	Type len_sq = _mm_shuffle_ps(sums, sums, _MM_SHUFFLE(0, 0, 0, 0));
 	// clang with '-ffast-math' (which you should not use!) can generate _mm_rsqrt_ps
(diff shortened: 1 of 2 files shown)


Jolt/Math/Vec3.inl
```diff
@@ -1091,12 +1010,14 @@ Vec3 Vec3::Normalized() const
...
-	__m128 shuf = _mm_movehdup_ps(mul);
-	__m128 sums = _mm_add_ps(mul, shuf);
-	shuf = _mm_movehl_ps(shuf, sums);
-	sums = _mm_add_ss(sums, shuf);
+	Type mul = _mm_mul_ps(mValue, mValue);
+	Type shuf = _mm_movehdup_ps(mul);
+	Type sums = _mm_add_ps(mul, shuf);
+	shuf = _mm_movehl_ps(mul, mul);
+#ifdef JPH_CROSS_PLATFORM_DETERMINISTIC
+	shuf = _mm_add_ps(_mm_setzero_ps(), shuf);
```
jrouwe, line 1013: Can you explain why this is needed? (I don't see it)

Jolt/Math/Vec3.inl
```diff
@@ -1230,6 +1151,33 @@ float Vec3::ReduceMax() const
...
+#elif defined(JPH_USE_NEON)
+	Type v = vsetq_lane_f32(0, mValue, 3);
+	return vaddvq_f32(v);
+#elif defined(JPH_USE_RVV)
+	const vfloat32m1_t zeros = __riscv_vfmv_v_f_f32m1(0.0f, 3);
+	const vfloat32m1_t v = __riscv_vle32_v_f32m1(mF32, 3);
+	const vfloat32m1_t sum = __riscv_vfredosum_vs_f32m1_f32m1(v, zeros, 3);
+	return __riscv_vfmv_f_s_f32m1_f32(sum);
+#else
+	float sum = 0.0f;
```
jrouwe, line 1169: We should make this:

```
return (mF32[0] + mF32[1]) + mF32[2];
```

(and this is probably why you needed `shuf = _mm_add_ps(_mm_setzero_ps(), shuf);`)

Jolt/Math/Vec3.inl
```diff
@@ -948,85 +921,44 @@ Vec4 Vec3::DotV4(Vec3Arg inV2) const
...
-	shuf = _mm_movehl_ps(shuf, sums);
-	sums = _mm_add_ss(sums, shuf);
-	return _mm_cvtss_f32(_mm_sqrt_ss(sums));
-#elif defined(JPH_USE_NEON)
-	float32x4_t mul = vmulq_f32(mValue, mValue);
-	mul = vsetq_lane_f32(0, mul, 3);
-	float32x2_t sum = vdup_n_f32(vaddvq_f32(mul));
-	return vget_lane_f32(vsqrt_f32(sum), 0);
+#if defined(JPH_USE_NEON)
+	// Minor differences in NaN handling make sqrt() produce
```
jrouwe: Is this still true now that I made a custom `Sqrt` function?

jrouwe (commented):
Note that there's still one test failing.

Jolt/Math/Vec3.inl
asenkew: Looks not needed, removed

jrouwe:
The macOS fail should be fixed by a7b3be487e74fc09e50be07cf2908208b24c0aad

jrouwe:
I finished your change in #2023. The reason why the test was failing is that MSVC was deciding that this code:

```
#ifdef JPH_CROSS_PLATFORM_DETERMINISTIC
	shuf = _mm_add_ps(_mm_setzero_ps(), shuf);
#endif
```

Wasn't worth emitting, so it was doing `(x + y) + z` instead of `(x + y) + (z + 0)` which returns a different result when x = y = z = -0.0f.

Thanks!

asenkew:
Great, thank you much!

PR #2018, opened 2026-05-13
MTLPipelineOptionBindingInfo was introduced in iOS 17 / macOS 14 SDK

jrouwe:
Fall back to the old name on older SDKs so the file still compiles with Xcode <= 15.x toolchains


jrouwe:
@zentia Are you ok with me taking this change into Jolt?

PR #2025, opened 2026-05-18
Adding options to Linux CMake build script

alkavan:
Hey.
I added JoltPhysics to the Magnum Graphics Engine, I'm making adjustment for Linux systems: [Adding Jolt Physics integration as JoltIntegration #121](https://github.com/mosra/magnum-integration/pull/121)

So I added some options so I can package both Release and Debug artifacts, for example as RPM, along with a shared library. That also aligned some of the conversions Magnum has and make sense. The default behavior of the script is kept. You can just run without parameters, defaults would be used, and no additional options are required.

Additional script options:
* Adding `--shared` to build `.so` instead
* Adding `--postfix-debug` to append `-d` to artifacts
* Adding `--env-script` to source a script before running build

I will gladly later add an RPM package spec for Fedora and RHEL systems using these options.

The Vulkan environment script was required for `dxc` on my Fedora system. There are some "unofficial" packages but the easiest way to get the compiler is actually just download whatever Vulkan SDK you want.

### DIFF
--- a/Build/cmake_linux_clang_gcc.sh
+++ b/Build/cmake_linux_clang_gcc.sh
@@ -1,28 +1,79 @@
 #!/bin/sh
 
-if [ -z $1 ]
-then
-	BUILD_TYPE=Debug
-else
-	BUILD_TYPE=$1
-	shift
+BUILD_TYPE=""
+COMPILER=""
+SHARED_LIB=false
+POSTFIX_DEBUG=false
+ENV_SCRIPT=""
+CMAKE_EXTRA_ARGS=""
+
+for arg in "$@"; do
+  case "$arg" in
+    --shared)
+      SHARED_LIB=true
+      ;;
+    --postfix-debug)
+      POSTFIX_DEBUG=true
+      ;;
+    --env-script=*)
+      ENV_SCRIPT="${arg#*=}"
+      ;;
+    *)
+      if [ -z "$BUILD_TYPE" ]; then
+        BUILD_TYPE="$arg"
+      elif [ -z "$COMPILER" ]; then
+        COMPILER="$arg"
+      else
+        CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS $arg"
+      fi
+      ;;
+  esac
+done
+
+if [ -z "$BUILD_TYPE" ]; then
+  echo "No build type argument, using default BUILD_TYPE=Debug"
+  BUILD_TYPE="Debug"
+fi
+if [ -z "$COMPILER" ]; then
+  echo "No compiler argument, using default COMPILER=clang++"
+  COMPILER="clang++"
 fi
 
-if [ -z $1 ]
-then
-	COMPILER=clang++
-else
-	COMPILER=$1
-	shift
+echo ""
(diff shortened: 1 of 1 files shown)


alkavan:
Found some pasted text that slipped in the previous commit.

jrouwe:
Don't know if you noticed, but CI is failing with your change.

> I will gladly later add an RPM package spec for Fedora and RHEL systems using these options.

Unfortunately I cannot support creating packages for these systems. There are too many build systems / os-es around and my time is limited. Also, the library has quite a few configuration options at the moment (e.g. double precision on/off), which makes distributing binary packages less useful. That said, the community has provided some of these (e.g. a vcpkg build) and I am linking to the ones I know about in the [Other Build Tools](https://github.com/jrouwe/JoltPhysics/tree/master/Build#other-build-tools) section. If you want to maintain a package, then I could link to it, but I don't want to make this part of the main repo as it then becomes my burden to maintain.

alkavan:
Yeah I saw this!
It's easy fix, `sh` /  `bash` compatibility issue.

I don't expect you to provide binaries packages, it makes sense. I planned on providing them myself. But I wanna make it slightly more easier for people who want to make their own packages. This can be useful not only for RPMs.

alkavan:
@jrouwe should work now, can you approve the workflow?

alkavan:
Some kind of GitHub issue?

```
Error: fatal: could not read Username for 'https://github.com/': terminal prompts disabled
Error: The process '/usr/bin/git' failed with exit code 128
```

alkavan:
Note to myself, fix this.
```
Run ./cmake_linux_clang_gcc.sh Distribution clang++-18 -DCROSS_COMPILE_ARM=ON -DCROSS_PLATFORM_DETERMINISTIC=ON -DCROSS_COMPILE_ARM_TARGET="" -DTARGET_VIEWER=OFF -DTARGET_SAMPLES=OFF -DTARGET_HELLO_WORLD=OFF -DTARGET_UNIT_TESTS=ON -DTARGET_PERFORMANCE_TEST=ON -DJPH_USE_VK=OFF -DCMAKE_CXX_FLAGS="--target=riscv64-unknown-linux-gnu -march=rv64gcv" -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld"

Usage: ./cmake_linux_clang_gcc.sh [BUILD_TYPE] [COMPILER] [--shared] [--postfix-debug] [--env-script=...]
Build type options: Debug, Release, Distribution, ReleaseUBSAN, ReleaseASAN, ReleaseTSAN, ReleaseCoverage
Compiler options: clang++, clang++-XX, g++, g++-XX where XX is the version

Generating Makefile for build type "Distribution" and compiler "clang++-18" in folder "Linux_Distribution"
CMake Error: Unknown argument -march=rv64gcv
CMake Error: Run 'cmake --help' for all supported options.

Compile by running:
cd Linux_Distribution && make -j4
Run tests: ./UnitTests
```

alkavan:
@jrouwe Could you run the workflow one more time? I tested as much as I could, hope it's fine now. Some of the issue is because it's `sh` and not `bash` I wanted to keep that compatibility. If this would fail I will probably just give up and just use it on my side

jrouwe:
Thanks, everything seems to work now!

PR #2048, opened 2026-06-09
fix assertion in CollideConvexVsTriangles::Collide

virt00l:
Penetration_axis initialized from the unnormalized triangle normal can be near-zero when an instance is scaled down (normal magnitude scales with area ~ scale^2) tripping JPH_ASSERT(!ioV.IsNearZero()) in GetPenetrationDepthStepGJK. sCollideConvexVsConvex already guards the same warm-start; this brings CollideConvexVsTriangles in line.

### DIFF
--- a/Jolt/Physics/Collision/CollideConvexVsTriangles.cpp
+++ b/Jolt/Physics/Collision/CollideConvexVsTriangles.cpp
@@ -69,6 +69,11 @@ void CollideConvexVsTriangles::Collide(Vec3Arg inV0, Vec3Arg inV1, Vec3Arg inV2,
 	// and the penetration axis is the shortest distance along to push B out of collision, we use the inverse of the triangle normal as an initial penetration axis. This has been seen
 	// to improve performance by approx. 5% over using a fixed axis like (1, 0, 0).
 	Vec3 penetration_axis = -triangle_normal, point1, point2;
+
+	// Ensure that we do not pass in a near zero penetration axis
+	if (penetration_axis.IsNearZero())
+		penetration_axis = Vec3::sAxisX();
+
 	EPAPenetrationDepth pen_depth;
 	EPAPenetrationDepth::EStatus status;
 


virt00l:
The actual assert happened on convex collision with tree which was scaled to ~5%

jrouwe:
Hello,

In the case of `sCollideConvexVsConvex`, it uses the delta between the center of masses as initial guess. Since you can't control the relative positions of bodies, we need this additional condition. 

When colliding with a triangle, it's a different story. If the `JPH_ASSERT(!ioV.IsNearZero())` assert triggers then it means that your triangle has an area of less than 1 mm^2 which I can't recommend. Even with your fix, you could run into floating point accuracy issues as it tries to compare a large shape with a tiny triangle. Triangles these small are also bad for performance as a convex object could overlap (and require testing against) many triangles at once. I would recommend passing your (non-scaled) triangle mesh through a mesh optimizer (e.g. https://github.com/zeux/meshoptimizer) to remove small triangles before handing the mesh to Jolt.

Note that this issue has come up several times before, and as far as I know, this solution has worked for the reporters so far.

virt00l:
> Hello,
> 
> In the case of `sCollideConvexVsConvex`, it uses the delta between the center of masses as initial guess. Since you can't control the relative positions of bodies, we need this additional condition.
> 
> When colliding with a triangle, it's a different story. If the `JPH_ASSERT(!ioV.IsNearZero())` assert triggers then it means that your triangle has an area of less than 1 mm^2 which I can't recommend. Even with your fix, you could run into floating point accuracy issues as it tries to compare a large shape with a tiny triangle. Triangles these small are also bad for performance as a convex object could overlap (and require testing against) many triangles at once. I would recommend passing your (non-scaled) triangle mesh through a mesh optimizer (e.g. https://github.com/zeux/meshoptimizer) to remove small triangles before handing the mesh to Jolt.
> 
> Note that this issue has come up several times before, and as far as I know, this solution has worked for the reporters so far.

Hmm, okay. I'll dig geometry of said tree a bit. 
But I believe that the issue there was due to ~5% scale (I specifically noted it in comment above) of mesh and not geometry itself.
Since then we limited scale floor to 1/16 (6.25%) so it's should be enough to mitigate this bug; although 5% scale is not something _too_ unusal/insane (I think).
Also we already using that (meshopt), I think, but worth re-checking ofc.

virt00l:
Hi,

I dug into the tree geometry like you suggested. First, a correction to my earlier message - the instance that crashed was actually at 1/32 scale (~3.1%), not the ~5% I quoted; I'd misread the value.

The mesh itself is clean - no degenerate or sliver triangles, the smallest face is around 96 mm^2 unscaled, nothing obviously broken. The real problem is that it's just far too dense for collision (1568
triangles for one tree), so once an instance gets scaled down the small faces fall under the limit. At that 1/32 scale a few dozen of them do, all near the trunk base.

So you're right that it's a small-triangle issue, and we'll fix it on our side - simplify the collision proxy and double-check meshopt actually runs on collision meshes and not only the visual LODs.

I'd still like to make the case for the guard though. On a dedicated server we place thousands of artist-scaled instances and can't fully guarantee none ever ends up under the limit - even our 1/16 scale
floor still leaves a couple of triangles tripping it on this particular tree. We'd much rather Jolt skip a degenerate triangle than assert, the same way CollideConvexVsConvex already does. And in release
builds the assert is compiled out, so instead of a clean skip we end up with a NaN normal and garbage depth, which is worse than a crash to track down. The guard covers both cases.

Attaching the raw unscaled collision mesh in case it's useful. Smallest face is ~96 mm^2 unscaled; at 1/32 a few dozen drop under 1 mm^2. Thanks for the meshopt pointer.

[tree_fagus_small_c.obj.zip](https://github.com/user-attachments/files/28969010/tree_fagus_small_c.obj.zip)

jrouwe:
The assert itself is pretty harmless. The code is protected so that even when you pass in ioV = (0, 0, 0) it doesn't divide by zero. So unless your assert handler actually forces a crash it should be fine. The main risk is that due to floating point precision you get the wrong contact point / penetration axis leading to a 'ghost collision' somewhere. I don't think you should get NaNs because of this. The assert mainly serves as a reminder that there is bad content.

> We'd much rather Jolt skip a degenerate triangle than assert, the same way CollideConvexVsConvex already does

The triangle is not skipped, only the first axis that is fed into the GJK collision detection algorithm changes. That shouldn't have a huge effect on the outcome, it affects how fast the (iterative) algorithm converges.

virt00l:
Thanks, that clears it up - good to know the zero case is handled and there's no NaN risk.

The fatal dump is actually on purpose for us. In our test/UAT builds we force a process dump on asserts like this as a content tripwire - internal QA can't realistically cover every instance of every
asset, so we want it loud there. So this was the "bad content reminder" working as intended on our end, just louder than you'd run it.

Which means no need for the guard - we'll fix it at the source instead (the tree's collision is heavily over-tessellated, and we've added a scale floor), and keep live servers from going fatal on it.
Appreciate you walking through the details.

PR #2063, opened 2026-06-24
docs(ProjectList): added Autodesk Synthesis

Dhruv-0-Arora:
https://github.com/autodesk/synthesis


jrouwe:
Thanks! Did you run into any issues while integrating?

azaleacolburn:
Hiii! Sorry for the late response, everything has been hectic for us for the past few weeks. Everything has been going fairly smoothly for us so far.

However, I do have some questions about memory management in JoltPhysics.JS. Would it be possible for us to chat sometime one-on-one? This could be over text or call, whatever works for you.

jrouwe:
Hello, I'd be happy to answer your questions. I would however prefer them to be asked [here](https://github.com/jrouwe/JoltPhysics.js/discussions). That makes it easier for me to track them and benefits others with similar questions as well.

PR #2095, opened 2026-08-08
Reduce sleep-stats wire sphere cost

glitziestudios:
BodyManager::Draw calls DrawWireSphere 3 times per body per step, using the default 3 subdivision levels when mDrawSleepStats is true. That amounts to 1536 line segments per sphere, for a total of 4608 per body. 

Needless to say, that's way too many for visualizing sleep test spheres, so I passed 0 InLevel to DrawWireSphere. 1 works fine too, but I didn't see much of a visual difference (from afar, anyway).


jrouwe:
Unfortunately I won't be near a computer in the next couple of weeks, so it will be a while before I can check this.

jrouwe:
Thanks!

PR #2118, opened 2026-09-07
Docs: document INTERPROCEDURAL_OPTIMIZATION for XCFramework builds

### LINKED ISSUE #1261: Release build for mac will enable bitcode
In Release/Distribution builds for macOS/iOS/visionOS, `INTERPROCEDURAL_OPTIMIZATION` is ON so bitcode is enabled.
I would like to see bitcode disabled by default as it has been deprecated since Xcode14.
Also, there are cases where bitcode is required for older models, so I think it would be better to be able to enable it as an option.

Probably related to the following issue and pull req.
https://github.com/jrouwe/JoltPhysics/issues/418
https://github.com/jrouwe/JoltPhysics/pull/357

zaveshaa:
Closes #1261

When a static Jolt library is bundled into an XCFramework, the `-flto=thin` bitcode embedded by `INTERPROCEDURAL_OPTIMIZATION` causes errors like `Unknown header: 0xb17c0de` on Apple platforms. Document this and how to turn the option off for XCFramework builds.

No code changes; documentation only.

### DIFF
--- a/Build/README.md
+++ b/Build/README.md
@@ -227,6 +227,10 @@ If you receive the following error when linking:
 
 Then you have not enabled interprocedural optimizations (link time optimizations) for your own application. See the INTERPROCEDURAL_OPTIMIZATION option in CMakeLists.txt.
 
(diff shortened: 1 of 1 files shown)


jrouwe:
Thanks!

PR #2119, opened 2026-09-07
Fix SetHeights corrupting the range block hierarchy when the block count is not a power of 2

spldm:
## Problem

`HeightFieldShape` stores its range block hierarchy with a stride of `1 << level` for every level except the most detailed one, which is clamped to `(num_blocks + 1) / 2` (constructor, "Create blocks" loop, and `WalkHeightField`: `stride = min(1U << level, max_stride)`).

`SetHeights` rebuilds the parent levels through `sGetRangeBlockOffsetAndStride(num_blocks >> 1, max_level - 1, ...)`, whose stride is `(n + 1) >> 1` at every level. For a block count that is a power of 2 the two agree. For any other block count they differ from the first parent level on (454 blocks: 114 vs 128, then 57 vs 64, ...), so the updated min/max values are written into unrelated nodes and the real parents keep their old range. Later iterations also read their source ranges with the wrong stride.

The effect is invisible to `CastRay` most of the time (a long ray's AABB still overlaps the wrong range) but `CollideShape` / `CharacterVirtual` queries with a small vertical extent get culled over the cells whose ancestor received a foreign range. In our game a `CharacterVirtual` walking over a 908 x 908 sample terrain (block size 2) fell through the ground at a fixed world position after the first `SetHeights` call elsewhere on the field, and only regained contact once it had sunk ~0.5 m, i.e. once its box reached the foreign range.

## Fix

Use the stored layout when updating the hierarchy: the destination level `max_level - 2` lives at `sGridOffsets[max_level - 2]` with stride `1 << (max_level - 2)`. The clamp to `(num_blocks + 1) / 2` never applies below the most detailed level because `1 << level <= 2^(max_level - 2) < (num_blocks + 1) / 2` for every `num_blocks > 2^(max_level - 1)`, so no `min()` is needed. The source stride for the next iteration follows automatically since it is taken from the previous destination.

## Test

`TestSetHeightsNonPowerOf2BlockCount` builds a 20 x 20 field with block size 2 (10 blocks per side), raises a block aligned rectangle in one corner with `SetHeights`, and then collides a small sphere with every untouched cell. Before the fix the cells under the node that received the raised range (samples x 16..19, z 8..11) report no hit; after the fix every cell collides and the raised part is found at its new height.

### DIFF
--- a/Docs/ReleaseNotes.md
+++ b/Docs/ReleaseNotes.md
@@ -10,6 +10,7 @@ For breaking API changes see [this document](https://github.com/jrouwe/JoltPhysi
 * If `GJKClosestPoint::CastShape` failed to converge, it was accidentally using the rejected support point to calculate the contact point. In some cases this could lead to returning the wrong contact point.
 * When colliding with very long and thin triangles, GJK could fail to find the closest point to the triangle because it used a fixed epsilon to determine if the triangle was too thin. Changed this epsilon so that it is dependent on the length of the edges of the triangle so that we fall back to closest point to edge in those cases.
 * MinGW32 defines `__STDCPP_DEFAULT_NEW_ALIGNMENT__` as 16 but its allocations are actually 8 byte aligned. This caused unaligned read access violations in `TriangleSplitterBinning::mBins`.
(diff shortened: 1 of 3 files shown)


jrouwe:
I've verified the change and it looks good to me. Thanks!

PR #2120, opened 2026-09-08
Fix Thread Performance Checker warning on macOS

zaveshaa:
The GCD semaphore used by the job system is associated with the default QoS class. When the main thread runs at the user interactive QoS class, the Thread Performance Checker reports a priority inversion when the main thread waits for the worker threads.

Reduce the QoS of the calling thread to the default class in `JobSystemThreadPool::Init` so the warning no longer triggers.

Fixes #1961

### DIFF
--- a/Jolt/Core/JobSystemThreadPool.cpp
+++ b/Jolt/Core/JobSystemThreadPool.cpp
@@ -11,6 +11,8 @@
 
 #ifdef JPH_PLATFORM_LINUX
 	#include <sys/prctl.h>
+#elif defined(JPH_PLATFORM_MACOS)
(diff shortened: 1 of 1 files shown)


jrouwe:
I don't think this is the correct fix. This means that anyone using the library will see their main thread's priority reduced to default QoS. I think it's bad manners for a library to mess with threads that it doesn't own.

zaveshaa:
Thanks for the feedback, you're right — Jolt shouldn't touch the QoS of a thread it doesn't own. Closing this.

zaveshaa:
Closed per maintainer feedback.

PR #2123, opened 2026-09-11
Use per-body contact points as friction lever arms (fixes #2121)

### LINKED ISSUE #2121: Friction impulses on speculative contacts use the midpoint of the two contact points, which adds `separation/2` to the lever arm and breaks angular-momentum conservation at the contact point
**Version**: Jolt Physics v5.3.0 (single-threaded job system, deterministic configuration)

Observed on v5.3.0 (the version pinned in our project); not yet re-run on v5.6.0.

**Summary**

For a speculative contact (the two shapes are still separated by `δ > 0` when the constraint is built), `WorldContactPoint::TemplatedCalculateFrictionAndNonPenetrationConstraintProperties` computes a single application point `p = 0.5 * (inWorldSpacePosition1 + inWorldSpacePosition2)` and uses the resulting `r1` / `r2` for **both** the non-penetration constraint **and** the two friction axes. Translating the application point along the contact normal is harmless for the normal axis (`(r + α·n) × n = r × n`), but it changes the friction axes (`(r + α·n) × t = r × t + α·(n × t)`). For a sphere resting/bouncing on a plane this makes the effective friction lever arm `R + δ/2` instead of `R`, so every sliding speculative impact injects a spurious angular impulse.

**Reproduction**

- Sphere: radius `R = 0.1 m`, mass `m = 1 kg`, solid-sphere inertia `I = (2/5)mR² = 0.004 kg·m²`.
- Initial state: centre at `z = 1.35 m`, linear velocity `(+2, 0, 0) m/s`, zero angular velocity.
- Static box floor, top face at `z = 0`, well clear of the edges.
- Materials: combined friction `μ = 0.5`, combined restitution `e = 0.8`; gravity `9.81 m/s²`.
- Stepping: `Update(dt = 50 ms, collision_steps = 10)` → `Δt_sub = 5 ms`; `mSpeculativeContactDistance = 0.02 m` (default); `mMinVelocityForRestitution = 0.140071 m/s`.
- Let the sphere bounce until it rolls without slipping, then measure the steady horizontal speed.

**Expected**

Both the normal impulse and the friction impulse act at the sphere's contact point, so the angular momentum about that point, `L = m·v_x·R + I·ω_y`, is conserved across every impact and every flight phase. With `L₀ = 0.2 kg·m²/s` the pure-rolling end state is `v_x = L₀ / (mR + I/R) = 5/7 · v₀ = 1.428571 m/s` (and `ω = 14.285714 rad/s`), independent of `μ`, of the solver iteration count, and of how many impacts the de-slipping takes.

**Actual**

`v_x = 1.4503364 m/s`, i.e. **+1.52 %**. The whole error appears at the first (large-slip) impact, where `L` jumps from `0.200000000` to `0.203026097 kg·m²/s` (+1.513 %); all subsequent impacts conserve `L` to within 0.02 pp.

**Analysis**

The narrow phase uses the sub-step start position, so the contact is created while the sphere is still `δ = 11.4875 mm` above the floor (`gap(n) = 1.25 − g·Δt_sub²·n(n+1)/2`, first value `≤ 0.02 m` at `n = 100`). With `mCombinedRestitution > 0` the whole bounce (incoming 4.95 m/s, outgoing 3.92 m/s) is applied in that single sub-step, and friction is applied in the same sub-step at the same point `p`. The closed form for the spurious angular momentum is

```
ΔL = |J_t| · δ / 2
```

With the measured tangential impulse `|J_t| = 0.527 N·s` this predicts `ΔL/L₀ = +1.513 %` against a measured `+1.513 %` (agreement to 0.02 %). Two independent consequences confirm the mechanism:

- Reducing `mSpeculativeContactDistance` to 0.01 m or 0.005 m makes the contact be detected one sub-step later, at `δ = −13.2775 mm` (i.e. penetrating), so the lever arm becomes `R − |δ|/2` and the error **changes sign**; both settings give the same `≈ −2.1 %` because they select the same sub-step.
- The error is independent of `μ` (the friction impulse is not Coulomb-clamped here: `μ·J_n = 4.44 N·s ≫ 0.527 N·s`), so it is not a friction-model issue — any tangential impulse on a speculative contact is affected.

**Suggested fix**

Derive the friction lever arms from each body's own contact point instead of the midpoint, leaving the non-penetration axis untouched (the normal axis is provably invariant under a translation along the normal, so this keeps the change bit-exact for frictionless contacts):

```cpp
// inside the `if (inSettings.mCombinedFriction > 0.0f)` block
Vec3 rf1 = Vec3(inWorldSpacePosition1 - inBody1.GetCenterOfMassPosition());
Vec3 rf2 = Vec3(inWorldSpacePosition2 - inBody2.GetCenterOfMassPosition());
// pass rf1 / rf2 to both friction AxisConstraintPart instances
```

A fuller variant uses the per-body contact points for `r1` / `r2` everywhere; on convex-vs-convex manifolds `p₁ − p₂` is parallel to the normal, so the normal Jacobian and effective mass are mathematically unchanged and only the (physically correct) tangential relative-velocity term differs.

**Effect of the suggested fix (measured)**

| Observable | Before | After |
|---|---|---|
| `ΔL/L₀` at the first impact (three drop heights) | +1.513 % | **−0.00027 %** |
| Steady rolling speed vs `5/7·v₀` | +1.52 % | **0.0 %** |
| `CONTACT_BEGIN` / `CONTACT_END` counts for the case above | 10 / 9 | **10 / 9 (unchanged)** |
| Byte-identical regression baselines (5 recorded cases) | — | **4 of 5 unchanged**; only the rolling case (μ > 0 with an off-normal lever arm) changes |

Contact detection, manifold reduction, the contact cache and warm starting are untouched by the change; only the friction lever arm moves.

AliveverCom:
Fixes #2121.

**Problem**

`ContactConstraint::CalculateFrictionConstraintProperties` derives the two friction lever arms `r1` / `r2` from `friction_point`, the average of the per-contact midpoints `0.5 * (p1 + p2)` that the callers store in `ws_contacts`. For a speculative contact the two shapes are still separated by `δ > 0` when the constraint is built, so on a curved body the effective friction lever arm is `R + δ/2` instead of `R`. Every sliding speculative impact then injects a spurious angular impulse `ΔL = |J_t|·δ/2` at the contact point (closed form and measurements in #2121: +1.51 % of the contact-point angular momentum on the first impact of a bouncing sphere, steady rolling speed +1.52 % off `5/7·v0`, independent of `μ`, restitution and solver iterations).

**Why only the friction axes**

Translating the application point along the contact normal leaves the non-penetration Jacobian unchanged (`(r + α·n) × n = r × n`), but changes the friction Jacobians (`(r + α·n) × t = r × t + α·(n × t)`). So the non-penetration constraint keeps using the shared midpoint and stays bit-exact; only the two friction `AxisConstraintPart`s get their lever arms from each body's own averaged contact points. Frictionless scenes are therefore unaffected.

**Change**

- The two callers keep both `p1_ws` and `p2_ws` per contact instead of only their midpoint.
- `CalculateFrictionConstraintProperties` averages the body-1 points and the body-2 points separately for `r1` / `r2`; the midpoint of the two averages is still used for the determinism log and for `mDistanceToFrictionCenter` (a normal-plane distance, invariant under the translation).
- Angular friction and the non-penetration constraints are untouched.
- `EstimateCollisionResponse.cpp` (the standalone response estimator) is updated in the same way so that it keeps matching the solver; `EstimateCollisionResponseTests` passes with both sides changed.

**Effect (measured on v5.3.0 with the equivalent 4-line change; this PR ports it to master's refactored friction setup)**

| Observable | Before | After |
|---|---|---|
| ΔL/L₀ at the first sliding impact (three drop heights) | +1.513 % | −0.00027 % |
| Steady rolling speed vs `5/7·v0` | +1.52 % | 0.0 % |
| Contact begin/end counts on the same scene | 10 / 9 | 10 / 9 |
| Byte-identical regression baselines (5 recorded scenes) | — | 4 of 5 unchanged; only the rolling scene (μ > 0, off-normal lever arm) changes |

Master builds with the change with no new warnings, and the unit tests pass on my machine (MinGW-w64 GCC 16, Windows, Release) except `ComputeTests::TestComputeDX12`, which fails on this machine without the change as well (no DX12 / dxc available).

Happy to adjust the approach (e.g. a per-body friction point only for speculative contacts, or a different placement of the averaging) if you prefer.

### DIFF
--- a/Jolt/Physics/Collision/EstimateCollisionResponse.cpp
+++ b/Jolt/Physics/Collision/EstimateCollisionResponse.cpp
@@ -74,6 +74,7 @@ void EstimateCollisionResponse(const Body &inBody1, const Body &inBody2, const C
 	ContactConstraintPart<EMotionType::Dynamic, EMotionType::Dynamic> contact_constraints[ContactPoints::Capacity];
 	Vec3 contact_points[ContactPoints::Capacity];
 	Vec3 friction_point = Vec3::sZero();
+	Vec3 friction_point1 = Vec3::sZero(), friction_point2 = Vec3::sZero();
 	for (uint c = 0; c < num_points; ++c)
 	{
 		// Calculate contact points relative to body 1 and 2
@@ -82,6 +83,12 @@ void EstimateCollisionResponse(const Body &inBody1, const Body &inBody2, const C
 		// Calculate friction point
 		contact_points[c] = p;
 		friction_point += p;
+
+		// Calculate friction lever arm points per body so that the friction lever arm is not perturbed
+		// by the normal-direction separation of a speculative contact (the non-penetration constraint
+		// above is unaffected by this separation and keeps using the midpoint p)
+		friction_point1 += inManifold.mRelativeContactPointsOn1[c];
+		friction_point2 += inManifold.mRelativeContactPointsOn2[c];
 
 		// Calculate contact point relative to com
 		Vec3 r1 = p - com1;
@@ -109,6 +116,8 @@ void EstimateCollisionResponse(const Body &inBody1, const Body &inBody2, const C
 	// Calculate distance to friction center for each point
 	float num_points_f = float(num_points);
 	friction_point /= num_points_f;
+	friction_point1 /= num_points_f;
+	friction_point2 /= num_points_f;
 	float distance_to_friction_center[ContactPoints::Capacity];
 	for (uint c = 0; c < num_points; ++c)
 	{
@@ -125,8 +134,8 @@ void EstimateCollisionResponse(const Body &inBody1, const Body &inBody2, const C
 	friction2.SetTotalLambda(0.0f);
 	if (inCombinedFriction > 0.0f)
 	{
-		Vec3 r1 = friction_point - com1;
-		Vec3 r2 = friction_point - com2;
+		Vec3 r1 = friction_point1 - com1;
+		Vec3 r2 = friction_point2 - com2;
 
 		friction1.CalculateConstraintProperties(inv_m1, inv_i1, r1, inv_m2, inv_i2, r2, outResult.mTangent1);
 		friction2.CalculateConstraintProperties(inv_m1, inv_i1, r1, inv_m2, inv_i2, r2, outResult.mTangent2);
--- a/Jolt/Physics/Constraints/ContactConstraintManager.cpp
+++ b/Jolt/Physics/Constraints/ContactConstraintManager.cpp
@@ -138,16 +138,28 @@ JPH_INLINE void ContactConstraintManager::WorldContactPoint<Type1, Type2>::Calcu
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 template <EMotionType Type1, EMotionType Type2>
-void ContactConstraintManager::ContactConstraint<Type1, Type2>::CalculateFrictionConstraintProperties(const Body &inBody1, const Body &inBody2, float inInvM1, float inInvM2, Mat44Arg inInvI1, Mat44Arg inInvI2, const RVec3 *inWorldSpaceContacts, Vec3Arg inWorldSpaceNormal, Vec3Arg inWorldSpaceTangent1, Vec3Arg inWorldSpaceTangent2, const ContactSettings &inSettings)
+void ContactConstraintManager::ContactConstraint<Type1, Type2>::CalculateFrictionConstraintProperties(const Body &inBody1, const Body &inBody2, float inInvM1, float inInvM2, Mat44Arg inInvI1, Mat44Arg inInvI2, const RVec3 *inWorldSpaceContacts1, const RVec3 *inWorldSpaceContacts2, Vec3Arg inWorldSpaceNormal, Vec3Arg inWorldSpaceTangent1, Vec3Arg inWorldSpaceTangent2, const ContactSettings &inSettings)
 {
 	// Calculate friction part
 	if (inSettings.mCombinedFriction > 0.0f)
 	{
-		// Calculate point where the friction applies by averaging the contact points
-		RVec3 friction_point = RVec3::sZero();
+		// Calculate the point where the friction applies for each body separately by averaging that
+		// body's own contact points. For a speculative contact the two bodies' contact points are still
+		// separated along the normal, so using their shared midpoint as the friction application point
+		// (as the non-penetration constraint does) would give both friction lever arms an error term along
+		// the normal. That term is fine for the non-penetration axis, but not for the friction axes, see below.
+		RVec3 friction_point1 = RVec3::sZero(), friction_point2 = RVec3::sZero();
 		for (uint32 i = 0; i < mNumContactPoints; ++i)
-			friction_point += inWorldSpaceContacts[i];
-		friction_point /= Real(mNumContactPoints);
+		{
+			friction_point1 += inWorldSpaceContacts1[i];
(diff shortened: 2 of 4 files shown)


jrouwe:
Thank you for the detailed description. I've added a unit test that replicates the setup that you describe. This confirms that the issue indeed exists, but it also highlights a problem with the proposed solution:

```C++
// This test checks friction vs speculative contact points. A ball bouncing on the floor will have speculative contact points.
// These speculative contact points should not be applied to a point outside the shape.
TEST_CASE("TestFrictionVsSpeculativeContactPoint")
{
	PhysicsTestContext c;
	BodyInterface &bi = c.GetBodyInterface();

	// Create floor
	Body &floor = c.CreateFloor();
	floor.SetRestitution(0.8f);
	floor.SetFriction(0.5f);

	// Create a sphere rolling over the floor
	const float radius = 0.1f;
	const float v0 = 2.0f;
	BodyCreationSettings bcs(new SphereShape(radius), RVec3(0, radius, 1), Quat::sIdentity(), EMotionType::Dynamic, Layers::MOVING);
	bcs.mRestitution = 0.8f;
	bcs.mFriction = 0.5f;
	bcs.mLinearVelocity = Vec3(v0, 0, 0);
	bcs.mLinearDamping = 0.0f;
	bcs.mAngularDamping = 0.0f;
	BodyID id1 = bi.CreateAndAddBody(bcs, EActivation::Activate);

	// Create a sphere initially bouncing on the floor and eventually rolling over the floor
	bcs.mPosition = RVec3(0, 1.35f, -1.0f);
	BodyID id2 = bi.CreateAndAddBody(bcs, EActivation::Activate);

	c.Simulate(5.0f);

	// Assuming the sphere will eventually roll without slipping so that w1 = v1 / r
	// and that the impulse at the friction point will lead to this through I * w1 = m * r * (v0 - v1)
	// with I = 2 / 5 * m * r^2 (inertia of a solid sphere) this leads to
	// 2 / 5 * m * r^2 * w1 = m * r * (v0 - v1)
	// 2 / 5 * r * w1 = v0 - v1
	// 2 / 5 * v1 = v0 - v1
	// v1 = 5 / 7 * v0
	float expected_v1 = (5.0f / 7.0f) * v0;
	float expected_w1 = expected_v1 / radius;

	// Test that both spheres have reached this velocity after some time
	Vec3 v1, w1, v2, w2;
	bi.GetLinearAndAngularVelocity(id1, v1, w1);
	bi.GetLinearAndAngularVelocity(id2, v2, w2);

	// ----> These succeed on current master, however they fail with this PR
	CHECK_APPROX_EQUAL(v1, Vec3(expected_v1, 0, 0), 1.0e-4f);
	CHECK_APPROX_EQUAL(w1, Vec3(0, 0, -expected_w1), 1.0e-3f);

	// ----> These fail with a large error on master but have a much smaller error with this PR
	CHECK_APPROX_EQUAL(v2, Vec3(expected_v1, 0, 0), 1.0e-4f);
	CHECK_APPROX_EQUAL(w2, Vec3(0, 0, -expected_w1), 1.0e-3f);
}
```

I've added annotations with `---->`. It boils down to that we're now erring on the other side: If there is a contact with penetration then we're applying the friction impulses on the surface of the sphere while the radius of the sphere is effectively smaller due to the penetration. The old approach was more accurate for that case. I guess we can detect this and switch to the old solution, but I'm not sure if that's the best approach.

PR #2130, opened 2026-09-17
Use JPH_NO_DEBUG instead of NDEBUG for JoltConfig.cmake

DarkDefender:
Previously, NDEBUG would be defined when you pulled in Jolt via the generated JoltConfig.cmake cmake files. This could lead to issues if the program you are linking it to is also checking and using NDEBUG.

I ran into this issue when trying to wire up Jolt within Blender. Because I built Jolt in "Release" mode, it would append the NDEBUG flag to all places in Blender where Jolt was defined as a dependency.

This would lead to segfaults as some parts of Blender would use NDEBUG specific variable types and some would not.


jrouwe:
I've created an alternative for this in #2132.

`NDEBUG` is defined by default by cmake when building in release mode (in `CMAKE_CXX_FLAGS_RELEASE`) and I don't want to remove that as some code in the MSVC runtime libraries responds to this. Instead, I've changed the code so that it no longer explicitly sets `NDEBUG` and relies on cmake to provide it by default. I'm assuming that Blender has overridden this default somehow (or else it would have had `NDEBUG` too).

jrouwe:
Let me know if the PR above works for you.

